PR #7, opened 2017-11-24
Add an example application with IdentityServer

jevakallio:
* Cleaned up the sample application
* Inited new React Native app
* Configured according to README

example-auth-app


jevakallio:
I'll add a `Example/README.md` and link to that from the main readme.

PR #14, opened 2018-01-02
Update build.gradle

ibaljeetvisions:
Required to remove 
"Error:Execution failed for task ':app:transformClassesWithJarMergingForDebug'.
> com.android.build.api.transform.TransformException: java.util.zip.ZipException: duplicate entry: android/support/v4/graphics/BitmapCompat.class"
Error while Building APK.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -35,5 +35,8 @@ repositories {
 
 dependencies {
     compile 'com.facebook.react:react-native:+'
-    compile "net.openid:appauth:0.7.0"
+    compile ("net.openid:appauth:0.7.0")
+    {
+       exclude module: 'support-compat'
+    }
 }


kadikraman:
Hi @ibaljeetvisions thanks for the PR!

I don't quite see why this change is needed though. I've ran the latest version of our library on the example app and it built without any errors. Would you be able to provide instructions on how I could replicate the error you're having so I could verify the fix?

ibaljeetvisions:
We are facing issue in release build in our React Native application.

Whenever we compile this library it shows an error message of duplicate 
record for some of the resources from Android's Support library v4.

We were using following libraries in our project

react-native-code-push

appcenter-analytics

appcenter-crashes

react-native-keychain

recat-native-authtool

and using App center for build generation. also it throws same error on 
Android studio too.

ibaljeetvisions:
We have created a sample application so that anyone can run it to reproduce the issue that we are facing from very long time.

The purpose of this sample code is to just provide the understandablity for the compile time error and enable everybody to debug it and find a correct solution.

It would be very baneficial for us if someone please fix the issue and update library accordingly.

**Given below is the link of the repository for SampleApp that we have just created:-**

https://github.com/ibaljeetvisions/SampleApp.git

**Given below is the error log that we were able to print when error occurred :-** 

`Error:Execution failed for task ':app:transformClassesWithJarMergingForDebug'. > com.android.build.api.transform.TransformException: java.util.zip.ZipException: duplicate entry: android/support/v4/widget/PopupMenuCompat.class`

kadikraman:
Hi @ibaljeetvisions - thanks for the sample application. It's been very helpful in debugging the issue.

Looks like you might have missed the setup instructions for Android in the [Readme](https://github.com/FormidableLabs/react-native-app-auth#android-setup) - easily done as the iOS instructions are so long!

I've opened a [PR on your example app](https://github.com/ibaljeetvisions/SampleApp/pull/1) to add the configuration changes so the project builds for me now.

Let me know if this works for you 👍

ibaljeetvisions:
Hi @kadikraman the solution suggested by you worked for us, Thanks for helping us to solve this as we were looking for this solution from very long.

ibaljeetvisions:
We are facing the below given crash while using this library for authentication.

However this crash is not happening in upper versions of android i.e versions greater that lollipop.

**Given below is the error log for the crash that we were able to track.**

```
@Override
    public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
        if (requestCode == 0)
        {

//        Here the if we print then--- data = null -> We are getting data = null here, However in upper versions of Android its able to provide data bundle
            if(data!=null)
            {
                AuthorizationResponse response = AuthorizationResponse.fromIntent(data);
                AuthorizationException exception = AuthorizationException.fromIntent(data);
                if (exception != null)
                {
                    promise.reject("RNAppAuth Error", "Failed to authenticate", exception);
                    return;
                }

                final Promise authorizePromise = this.promise;

                AuthorizationService authService = new AuthorizationService(this.reactContext);

                authService.performTokenRequest(
                        response.createTokenExchangeRequest(),
                        new AuthorizationService.TokenResponseCallback() {

                            @Override
                            public void onTokenRequestCompleted(
                                    TokenResponse resp, AuthorizationException ex) {
                                if (resp != null) {
                                    WritableMap map = tokenResponseToMap(resp);
                                    authorizePromise.resolve(map);
                                } else {
                                    promise.reject("RNAppAuth Error", "Failed exchange token", ex);
                                }
                            }
                        });
            }
            else
            {
//                Toast.makeText(reactContext, "Something went wrong!", Toast.LENGTH_SHORT).show();
            }
        }
    }
```
When we enter in onActivityResult() method after login, its not able deliver the result in  “ Intent data” in lower versions of Android.

It would be very helpful for us if someone please look into this error and suggest us an appropriate solution for this.

Please suggest a solution if you are familiar with this issue and got any solution for this.

Thanks

kadikraman:
I've investigated this and looks like Android 21 and up are fine, however earlier versions will fail, because the library is using `startActivityForResult` and `onActivityResult` which do not work in earlier versions.

According to [this](https://github.com/openid/AppAuth-Android/issues/299) and [this](https://github.com/openid/AppAuth-Android/issues/306) issue in the AppAuth-Android repo, the only way to fix this is to use `performAuthorizationRequest(request, completionIntent, cancelIntent)` directly instead of `startActivityForResult`.

I will close this PR now (as it was for an unrelated issue initially) and create a new issue to investigate this further. 🌟

Issue #30, opened 2018-02-08
Ability to pass "acr_values" OpenID Connect auth request parameter


sattaman:
I am currently migrating our app from oidc-client-js and implicit flow to app auth.

A feature we are currently using is passing acr_values when logging in directly to Facebook and Google.

`oidc.signinPopup({ acr_values: 'idp:Google' })`

It looks like this feature is supported by appAuth-iOS https://github.com/openid/AppAuth-iOS/issues/199

kadikraman:
Nice, so to confirm, the feature request is to be able to pass in anything to `additionalParameters`, in this case
```js
additionalParameters = {
  acr_values: 'idp:Google'
}
```

sattaman:
yes please

kadikraman:
Got a PR for this: https://github.com/FormidableLabs/react-native-app-auth/pull/33

Do you think it is ever necessary to have different `additionalParameters` for `authorize` vs `refresh`? I've made it so they're always the same at the moment, but could instead pass them directly to `authorize()` and `refresh()` functions 🤔

sattaman:
Wow that was fast! The only thing i'd flag up is i don't always pass the same parameters. So as it stands I'd need to have a new instance of appAuth per login. I don't need to pass parameters on the refresh call, I'm not sure whether people may need to have different params there or not.

kadikraman:
Good point. I've been thinking for a little while that the instance-based API isn't really ideal as it imposes restrictions that have no business existing. Perhaps it would be better to have a more functional API. I'll write down some ideas.

kadikraman:
@sattaman just published`v2.0.0` and we've changed the API to be much more flexible.

So you can do e.g.

```js
const config = '<your shared config without additionalParameters>';

// authenticate
const result = await authorize({ ...config, additionalParameters: {
  acr_values: 'idp:Google'
});

// refresh
const result = await refresh(config, {
  refreshToken: result.refreshToken,
});

```
Here are the [release notes](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v2.0.0) 🚀

sattaman:
I noticed , i refreshed you page and it had all changed , i actually checked you pr history to sanity check haha . looks good, updating now.

PR #33, opened 2018-02-08
Feature/additional parameters

kadikraman:
Fixes https://github.com/FormidableLabs/react-native-app-auth/issues/30

This will enable passing additional parameters to the auth server in the auth request.
They will be passed into the auth object in JavaScript

e.g.

```js
const appAuth = new AppAuth({
  issuer: '<YOUR_ISSUER_URL>',
  clientId: '<YOUR_CLIENT_ID>',
  redirectUrl: '<YOUR_REDIRECT_URL>',
  additionalParameters: { hello: 'world' },
});
```

- [x] implement `additionalParameters` on iOS
- [x] implement `additionalParameters` on Android
- [x] add documentation for the `config` object

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,9 @@ This is your configuration object for the client
 - **issuer**: (`string`) *REQUIRED* the url of the auth server
 - **clientId**: (`string`) *REQUIRED* your client id on the auth server
 - **redirectUrl**: (`string`) *REQUIRED* the url that links back to your app with the auth code
-- **additionalParameters**: (`object` | `null`) additional parameters that will be passed in the authorization request
+- **additionalParameters**: (`object` | `null`) additional parameters that will be passed in the authorization request.
+Must be string values! E.g. setting `additionalParameters: { hello: 'world', foo: 'bar' }` would add
+`hello=world&foo=bar` to the authorization request.
 
 ### `refresh`
 
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ const result = await appAuth.authorize(scopes);
 // returns accessToken, accessTokenExpirationDate and refreshToken
 ```
 
-## config
+#### config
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -54,6 +54,14 @@ const result = await appAuth.authorize(scopes);
 // returns accessToken, accessTokenExpirationDate and refreshToken
 ```
 
+#### config
+
+This is your configuration object for the client
+- **issuer**: (`string`) *REQUIRED* the url of the auth server
+- **clientId**: (`string`) *REQUIRED* your client id on the auth server
+- **redirectUrl**: (`string`) *REQUIRED* the url that links back to your app with the auth code
+- **additionalParameters**: (`object` | `null`) additional parameters that will be passed in the authorization request
```
jevakallio: Would it be helpful to document how these additional parameters are passed in the request or is that obvious to anyone who is using this?

android/src/main/java/com/reactlibrary/RNAppAuthModule.java
```diff
@@ -63,8 +66,29 @@ private WritableMap tokenResponseToMap(TokenResponse response) {
         return map;
     }
 
+    private HashMap<String, String> additionalParametersToMap(ReadableMap additionalParameters) {
+
+        HashMap<String, String> additionalParametersHash = new HashMap<>();
+
+        ReadableMapKeySetIterator iterator = additionalParameters.keySetIterator();
+
+        if (iterator.hasNextKey()) {
```
jevakallio: Lol

README.md
kadikraman: Good point. Essentially

```
{ hello: 'world` }
```

would be sent as

```
<url-and-other-params>&hello=world
```

But I'll make it more explicit in the docs.

Issue #17, opened 2018-01-29
idToken is missing on Android


dolly22:
idToken is missing in TokenResponse handling of authorize() on Android (the same mapping is used for authorize() and refresh())

https://github.com/FormidableLabs/react-native-app-auth/blob/efd5f7bae611f2b33a490eb7f95e89d9a017cf3b/android/src/main/java/com/reactlibrary/RNAppAuthModule.java#L59-L62

IOS uses different mapping for authorize() ...

https://github.com/FormidableLabs/react-native-app-auth/blob/efd5f7bae611f2b33a490eb7f95e89d9a017cf3b/ios/RNAppAuth.m#L57-L63

and refresh() ...

https://github.com/FormidableLabs/react-native-app-auth/blob/efd5f7bae611f2b33a490eb7f95e89d9a017cf3b/ios/RNAppAuth.m#L117-L121

PR #34, opened 2018-02-08
Bugfix/missing params

kadikraman:
Fixes https://github.com/FormidableLabs/react-native-app-auth/issues/17

This will add missing parameters to the token response on iOS and Android.
Also adding documentation on which variables are returned.


jevakallio:
LGTM!

Issue #35, opened 2018-02-09
Proposal: v2.0.0


kadikraman:
### Currently API:

```js
import AppAuth from 'react-native-app-auth';

const refreshTokenConfig = {
  issuer: <ISSUER>,
  redirectUrl: <REDIRECT_URL>,
  clientId: <CLIENT_ID>,
  additionalParameters: <ADDITIONAL_PARAMETERS>
};
const appAuth = new AppAuth(config);

// authorize
const scopes = [SCOPES];
const result = await appAuth.authorize(scopes);

// refresh
const scopes = [SCOPES];
const result = await appAuth.refresh(scopes);

// revoke
const tokenToRevoke = <TOKEN>;
const sendClientId = BOOLEAN;
const result = await appAuth.revokeToken(tokenToRevoke, sendClientId = false);
```

### Proposed API:

```js
import { authorize, refresh, revokeToken } from 'react-native-app-auth';

// authorize
const authorizeConfig = {
  issuer: <ISSUER>,
  redirectUrl: <REDIRECT_URL>,
  clientId: <CLIENT_ID>,
  scopes: [SCOPES],
  additionalParameters: <ADDITIONAL_PARAMETERS>
};
const result = await authorize(authorizeConfig);

// refresh
const refreshTokenConfig = {
  issuer: <ISSUER>,
  redirectUrl: <REDIRECT_URL>,
  clientId: <CLIENT_ID>,
  refreshToken: <TOKEN>,
  scopes: [SCOPES],
  additionalParameters: <ADDITIONAL_PARAMETERS>
};
const result = await refresh(refreshTokenConfig);

// revoke
const revokeTokenConfig = {
  issuer: <ISSUER>,
  clientId: <CLIENT_ID>,
  tokenToRevoke: <TOKEN>,
  sendClientId: BOOLEAN
};
const result = await revokeToken(revokeTokenConfig);
```

#### Pros
- API is easier to grasp, as it gets rid of the extra step of creating a new instance of AppAuth
- Easier to extend: the config is just an object

#### Cons
- There is more repetition
- Easier to lose track of shared config
- Config validation has to be done in every exposed function rather than once when instantiating

jevakallio:
How about allowing sharing the configuration for all three methods:
```
import { authorize, refresh, revokeToken } from 'react-native-app-auth';

// authorize
const baseConfig = {
  issuer: <ISSUER>,
  redirectUrl: <REDIRECT_URL>,
[...]

PR #37, opened 2018-02-09
Feature/new api

kadikraman:
- [x] update the JavaScript API as discussed in https://github.com/FormidableLabs/react-native-app-auth/issues/35
- [x] update tests
- [x] update documentation
- [x] update Example app

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,9 +104,8 @@ const config = {
   scopes: '<YOUR_SCOPES_ARRAY>',
 };
 
-const result = await refresh({
-  ...config,
-  refreshToken: `<REFRESH_TOKEN>`,
+const result = await refresh(config, {
+  refreshToken: `<REFRESH_TOKEN>`
 });
 ```
 
@@ -124,8 +123,7 @@ const config = {
   scopes: '<YOUR_SCOPES_ARRAY>',
 };
 
-const result = await revoke({
-  ...config,
+const result = await revoke(config, {
   tokenToRevoke: `<TOKEN_TO_REVOKE>`
 });
 ```
@@ -379,8 +377,7 @@ const refreshedState = await refresh({
 });
 
 // Revoke token, note that Identity Server expects a client id on revoke
-await revoke({
-  ...config,
+await revoke(config, {
   tokenToRevoke: refreshedState.refreshToken,
   sendClientId: true
 });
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -47,50 +49,81 @@ flow and returns the access token, refresh token and access token expiry date wh
...
+const config = {
+  issuer: '<YOUR_ISSUER_URL>',
+  clientId: '<YOUR_CLIENT_ID>',
+  redirectUrl: '<YOUR_REDIRECT_URL>',
+  scopes: '<YOUR_SCOPES_ARRAY>',
+};
+
+const result = await revoke({
+  ...config,
+  tokenToRevoke: `<TOKEN_TO_REVOKE>`
```
kitten: I'm not sure whether this is a sensible API design, as it's not part of a config.

`config` is currently pretty agnostic to how it's used, but `tokenToRevoke` is a one off parameter. Since `revoke` will always affect a single token maybe the functions should be changed to either:

`revoke: (config: {| ... |}, token: string) => Promise<>`

or:

`revoke: (config: {| ... |}) => (token: string) => Promise<>`

same could apply to the `refreshToken` above, since it changes when calling `authorize`. I'm not sure what `authorize: (...) => Promise<T>` resolves to. Maybe `revoke` and `refresh` should just accept the result of `authorize`?

kadikraman: I think the first suggestion makes sense, but with the second argument also being an object:

```js
revoke: (config: {| ... |}, { token: string }) => Promise<>
```

README.md
```diff
@@ -1,8 +1,10 @@
-[![Build Status](https://travis-ci.org/kadikraman/draftjs-md-converter.svg?branch=master)](https://travis-ci.org/kadikraman/draftjs-md-converter)
+[![Build Status](https://travis-ci.org/FormidableLabs/react-native-app-auth.svg?branch=master)](https://travis-ci.org/FormidableLabs/react-native-app-auth)
 [![npm version](https://badge.fury.io/js/react-native-app-auth.svg)](https://badge.fury.io/js/react-native-app-auth)
 
 # React Native App Auth
 
+### This is the API documentation for `react-native-app-auth >= 2.0.` See version `1.x` documentation [here](https://github.com/FormidableLabs/react-native-app-auth/tree/v1.0.1).
```
jevakallio: Can you make the link text the whole "See version `1.x` documentation here" sentence for accessibility reasons?

Issue #46, opened 2018-02-18
Better error handling for canceled auth flow


PeterKottas:
Report type:
BUG

Steps to reproduce:
Start current example app. 
Click authorize.
Not abandon authorization (click back)
Observe the red screen of death.

Expected result:
Exception should be thrown, catched and handled via user code. The code in example app indeed is surrounded with try catch block but it appears it's no use for this scenario.

Actual result:
App shows red screen of death 
![canceledauthflow](https://user-images.githubusercontent.com/10601911/36355353-2870e62c-14e2-11e8-8501-c2f7b9dcab4e.jpg)

kadikraman:
@PeterKottas that's because the example app uses `console.error` to show you the error rather than `console.log`. Replace it with `console.log` and it will fail silently. This was intentional to draw the user's attention to something going wrong. It will only happen during development.

I do agree that the red screen of death can be a bit distressful so we could draw the user's attention to the errors with an Alert instead. 😄

PeterKottas:
Damn you're absolutely right! Sorry for false alarm, I just saw red screen and figured all hell broke lose :D Still new to native, you learn every day, thanks!

PR #48, opened 2018-02-19
Use less scary error handling in the example app

kadikraman:
Resolves https://github.com/FormidableLabs/react-native-app-auth/issues/46

Show an alert instead of the red screen of death in the example app if the authentication fails.


jevakallio:
Retroactive review, this is all good. 

Sake of "in case you missed", `alert` is also polyfilled globally, so you can just do `alert("")` instead of importing `Alert.alert`. But no preference, no need to change.

PR #49, opened 2018-02-19
Add typescript definition

### LINKED ISSUE #47: Typescript definitions
I have created Typescript definitions, and wanted to check if you wanted to include this in the package or if I should simply add them to DefinitelyTyped.

```ts
declare module "react-native-app-auth" {
  interface IAppAuthProperties {
    scopes: string[];
    issuer: string;
    clientId: string;
    redirectUrl: string;
    aditionalParameters?: object;
  }

  function authorize(
    properties: IAppAuthProperties
  ): Promise<{
    accessToken: string;
    accessTokenExpirationDate: string;
    additionalParameters: object;
    idToken: string;
    refreshToken: string;
    tokenType: string;
  }>;

  function refresh(
    properties: IAppAuthProperties,
    { refreshToken: string }
  ): Promise<AuthorizeResult>;

  function revoke(
    properties: IAppAuthProperties,
    { tokenToRevoke: string, sendClientId: boolean }
  ): Promise<void>;
}
```

janpieterz:
Fixed #47 

First time adding definitions to a package, worked locally. Let me know if you want anything changed.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,9 +1,17 @@
-export interface AppAuthInput {
+export interface AuthConfiguration extends BaseAuthConfiguration {
     scopes: string[];
-    issuer: string;
-    clientId: string;
     redirectUrl: string;
     aditionalParameters?: object;
+  }
+
+  export interface BaseAuthConfiguration{
+    issuer: string;
+    clientId: string;
+  }
+
+  export interface RevokeConfiguration{
+    clientId: string;
+    issuer: string;
   }
 
   export interface AuthorizeResult {
@@ -15,14 +23,19 @@ export interface AppAuthInput {
     tokenType: string;
   }
 
-  export function authorize(properties: AppAuthInput): Promise<AuthorizeResult>;
+  export interface RevokeOptions {
+    tokenToRevoke: string;
+    sendClientId?: boolean;
+  }
+
+  export function authorize(properties: AuthConfiguration): Promise<AuthorizeResult>;
 
   export function refresh(
-    properties: AppAuthInput,
+    properties: AuthConfiguration,
     { refreshToken: string }
   ): Promise<AuthorizeResult>;
 
   export function revoke(
-    properties: AppAuthInput,
-    { tokenToRevoke: string, sendClientId: boolean }
+    properties: BaseAuthConfiguration,
+    options: RevokeOptions
 ): Promise<void>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,39 +1,44 @@
+export interface BaseAuthConfiguration {
+  issuer: string;
+  clientId: string;
+}
 export interface AuthConfiguration extends BaseAuthConfiguration {
-    scopes: string[];
-    redirectUrl: string;
-    aditionalParameters?: {[name: string]: string};
-  }
+  scopes: string[];
+  redirectUrl: string;
+  additionalParameters?: { [name: string]: string };
+}
 
-  export interface BaseAuthConfiguration{
-    issuer: string;
-    clientId: string;
-  }
+export interface RevokeConfiguration {
+  clientId: string;
+  issuer: string;
+}
 
-  export interface RevokeConfiguration{
-    clientId: string;
-    issuer: string;
-  }
+export interface AuthorizeResult {
+  accessToken: string;
+  accessTokenExpirationDate: string;
+  additionalParameters?: { [name: string]: string };
+  idToken: string;
+  refreshToken: string;
+  tokenType: string;
+}
 
-  export interface AuthorizeResult {
-    accessToken: string;
-    accessTokenExpirationDate: string;
-    aditionalParameters?: {[name: string]: string};
-    idToken: string;
-    refreshToken: string;
-    tokenType: string;
-  }
+export interface RevokeConfiguration {
+  tokenToRevoke: string;
+  sendClientId?: boolean;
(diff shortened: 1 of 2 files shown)


jevakallio:
@kitten can you take a peek from TS good practices point of view when you have a moment. 

@kadikraman These seem legit to me, though I'd check if the optionality of fields is correct - IIRC the `revoke` call actually requires fewer parameters in the config?

Only thing I'd change is the name `AppAuthInput` to something like `AuthConfiguration`.

index.d.ts
```diff
@@ -0,0 +1,28 @@
...
+  export function authorize(properties: AppAuthInput): Promise<AuthorizeResult>;
+
+  export function refresh(
+    properties: AppAuthInput,
+    { refreshToken: string }
+  ): Promise<AuthorizeResult>;
+
+  export function revoke(
+    properties: AppAuthInput,
+    { tokenToRevoke: string, sendClientId: boolean }
```
kadikraman: - The only properties needed for `revoke` are `clientId` and `issuer`. It doesn't matter if you pass in the whole configuration object, but it's not all used. I'm not sure if there's an easy way to reflect that in TypeScript?
- `sendClientId` is optional, it defaults to `false`

janpieterz: Thanks for the feedback!
As a proposal I made the `clientId` and `issues` in a separate interface that the `AuthConfiguration` extends, making it 1 definition. 
If we don't extend it'll mean we'd need to have two instances of the config object.

`sendClientId` made optional

janpieterz:
Any idea for a name ( 😄 ) for the `clientId` and `issuer` revoke configuration different than `BaseAuthConfiguration`?

Changed name to `AuthConfiguration` as well.

index.d.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  export interface RefreshConfiguration {
+    refreshToken: string;
+  }
+
+  export function authorize(config: AuthConfiguration): Promise<AuthorizeResult>;
+
+  export function refresh(config: AuthConfiguration, refreshConfig: RefreshConfiguration): Promise<AuthorizeResult>;
+
+  export function revoke(config: BaseAuthConfiguration, refreshConfig: RevokeConfiguration): Promise<void>;
```
kadikraman: Surely
```
revokeConfig: RevokeConfiguration
```

index.d.ts
```diff
@@ -0,0 +1,39 @@
+export interface AuthConfiguration extends BaseAuthConfiguration {
+    scopes: string[];
+    redirectUrl: string;
+    aditionalParameters?: {[name: string]: any};
```
kadikraman: Could you do
```
aditionalParameters?: {[name: string]: string};
```
as the values can technically only be strings?

index.d.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  export interface RevokeConfiguration{
+    clientId: string;
+    issuer: string;
+  }
+
+  export interface AuthorizeResult {
+    accessToken: string;
+    accessTokenExpirationDate: string;
+    aditionalParameters?: {[name: string]: any};
```
kadikraman: Same here

janpieterz:
Done, cheers!

index.d.ts
```diff
@@ -0,0 +1,39 @@
+export interface AuthConfiguration extends BaseAuthConfiguration {
```
kitten: Minor nit here, the indentation seems to be off

index.d.ts
```diff
@@ -0,0 +1,39 @@
...
+    scopes: string[];
+    redirectUrl: string;
+    aditionalParameters?: {[name: string]: string};
+  }
+
+  export interface BaseAuthConfiguration{
```
kitten: can you move this to the top as `AuthConfiguration` extends it before it's being defined. Just for readability :+1:

index.d.ts
```diff
@@ -0,0 +1,39 @@
+export interface AuthConfiguration extends BaseAuthConfiguration {
+    scopes: string[];
+    redirectUrl: string;
+    aditionalParameters?: {[name: string]: string};
```
kitten: small typo: a**dd**itionalParameters

index.d.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  export interface BaseAuthConfiguration{
+    issuer: string;
+    clientId: string;
+  }
+
+  export interface RevokeConfiguration{
```
kitten: can we make sure that there's a space before `{`?

@kadikraman: Can you point prettier—I think it's set up with lint-staged, right?—at all TS files as well?

index.d.ts
```diff
@@ -0,0 +1,39 @@
...
+    clientId: string;
+    issuer: string;
+  }
+
+  export interface AuthorizeResult {
+    accessToken: string;
+    accessTokenExpirationDate: string;
+    aditionalParameters?: {[name: string]: string};
```
kitten: same typo

janpieterz:
Had disabled prettier locally because it kept massively reformatting the package.json on edit. 

Now uses prettier formatting and added *.ts to lint-staged too.

PeterKottas:
Looks good to me! Gj

jevakallio:
Looking good, landed this to master, will push it out in the next release. Thanks @janpieterz @kitten @kadikraman!

PR #50, opened 2018-02-20
Allow insecure HTTP requests on Android

### LINKED ISSUE #45: Better error handling for required SSL
Report type:
BUG

Summary:
SSL appears to be enforced by AppAuth-Android library therefore you cannot use http for your idp. Makes sense but remember this limitation exists for localhost as well which is quite a letdown for dev and testing. 

Expected result:
If you try to connect to NON-SSL idp, a meaningful error message should be presented to the developer. 

Actual result:
App crashes with "Unfortunately, XXX has stopped. OK"

jevakallio:
Adds a new optional `dangerouslyAllowInsecureHttpRequests` flag that allows connecting to plain HTTP resources and HTTPS resources with self-signed or invalid certificates.

Fixes #45.

### DIFF
--- a/Example/ios/AppAuthExample/Info.plist
+++ b/Example/ios/AppAuthExample/Info.plist
@@ -39,6 +39,11 @@
 	<dict>
 		<key>NSExceptionDomains</key>
 		<dict>
+			<key>demo.identityserver.io</key>
+			<dict>
+				<key>NSExceptionAllowsInsecureHTTPLoads</key>
+				<true/>
+			</dict>
 			<key>localhost</key>
 			<dict>
 				<key>NSExceptionAllowsInsecureHTTPLoads</key>
--- a/android/src/main/java/com/reactlibrary/RNAppAuthModule.java
+++ b/android/src/main/java/com/reactlibrary/RNAppAuthModule.java
@@ -4,6 +4,7 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
+import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import com.facebook.react.bridge.ActivityEventListener;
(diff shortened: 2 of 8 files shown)


android/src/main/java/com/reactlibrary/utils/UnsafeConnectionBuilder.java
```diff
@@ -0,0 +1,128 @@
+package com.reactlibrary.utils;
```
jevakallio: This file is copied as-is from the AppAuth Android example application. Only modification here is that I changed the name from `ConnectionBuilderForTesting` to `UnsafeConnectionBuilder` https://github.com/openid/AppAuth-Android/blob/29329dce430edebdc80e9635895358e20971f286/app/java/net/openid/appauthdemo/ConnectionBuilderForTesting.java

PeterKottas:
Good stuff. Java is one language I hoped I'll never have to start using but it seems it's fairly easy to follow. Thanks again for the effort that went into this. I am curious, is this something you use in your projects? Or are you making the lib to make https://formidable.com/ bit more visible in the community? Reason I am asking is you are doing this admirably professionally. So it doesn't seem like your usual hobby lib :)

android/src/main/java/com/reactlibrary/RNAppAuthModule.java
```diff
@@ -96,24 +109,53 @@ private WritableMap tokenResponseToMap(TokenResponse response) {
...
+
+    private ConnectionBuilder createConnectionBuilder(Boolean allowInsecureConnections) {
+        if (allowInsecureConnections.equals(true)) {
+            return UnsafeConnectionBuilder.INSTANCE;
+        }
+
+        return DefaultConnectionBuilder.INSTANCE;
+    }
+
+    private Uri buildConfigurationUriFromIssuer(Uri openIdConnectIssuerUri) {
```
jevakallio: Since we are now using `fetchFromUrl` instead of `fetchFromIssuer`, we need to construct the issuer WKID endpoint ourselves. Annoyingly the method on `AuthorizationServiceConfiguration` was private, so I replicated it here. This should be a very stable method, since the URL format is driven by spec and not likely to change.

jevakallio:
> I am curious, is this something you use in your projects? Or are you making the lib to make formidable.com bit more visible in the community? Reason I am asking is you are doing this admirably professionally. So it doesn't seem like your usual hobby lib :)

@PeterKottas Hah, good question. The library was initially built for a client project. We are maintaining and developing it further pro-bono because we think it will be useful for other developers in the community, and when we'll next need to implement enterprise OAuth, we'll have a well-maintained and familiar tool to reach out for.

PeterKottas:
Nice one! 👍 Best of luck to you guys. I'll be using this a lot so I'll let you know if I find any more issues. Cheers.

Example/ios/AppAuthExample/Info.plist
```diff
@@ -39,6 +39,11 @@
 	<dict>
 		<key>NSExceptionDomains</key>
 		<dict>
+			<key>demo.identityserver.io</key>
```
kadikraman: Is this what need to be done on iOS in order to support insecure connections? If so, should we add a comment to the readme on it?

jevakallio: Yeah, could do that. The library already gives you a very informative warning saying the App Transport Security settings needs to be changed to allow HTTP connections. But no harm putting it in the README to save folks some cycles. Will amend 👍

PR #60, opened 2018-02-22
Typings fix

PeterKottas:
Just fixed errors typescript was complaining about.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,13 +17,13 @@ export type BaseAuthConfiguration =
       serviceConfiguration?: ServiceConfiguration;
     };
 
-export interface AuthConfiguration extends BaseAuthConfiguration {
+export type AuthConfiguration = BaseAuthConfiguration & {
   clientSecret?: string;
   scopes: string[];
   redirectUrl: string;
   additionalParameters?: { [name: string]: string };
   dangerouslyAllowInsecureHttpRequests?: boolean;
-}
+};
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -6,7 +6,7 @@ export interface ServiceConfiguration {
 }
 
 export type BaseAuthConfiguration =
-  | {
```
kitten: I'm pretty sure this was a style choice that @jevakallio introduced using an automated prettier format. Have you accidentally skipped our lint-staged formatting? https://github.com/FormidableLabs/react-native-app-auth/blob/master/package.json#L18-L19

index.d.ts
```diff
@@ -17,13 +17,13 @@ export type BaseAuthConfiguration =
       serviceConfiguration?: ServiceConfiguration;
     };
 
-export interface AuthConfiguration extends BaseAuthConfiguration {
+export type AuthConfiguration = BaseAuthConfiguration & {
```
kitten: This looks good to me. Good catch!

index.d.ts
PeterKottas: Ah ok, I see. Never seen it written like that, it reads like "nothing or this or that" to me so I was afraid if it wouldn't accept nothing the way it was. Hope that makes sense :D Anyways, I've reverted it. Rest should be good to go in.

PeterKottas:
👍

PR #71, opened 2018-03-14
Add support for Android API 16-19

### LINKED ISSUE #29: Android: support for API 16 - 19
Related issue: https://github.com/FormidableLabs/react-native-app-auth/pull/14

The current implementation supports **Android API 21 (Lollipop)** and above, however the parent library supports [Android API 16 (Jellybean)](https://github.com/openid/AppAuth-Android#requirements) and above.

This is because we are using the `startActivityForResult` and `onActivityResult` which do not work in earlier versions: when calling `startActivityForResult`, the callback to `onActivityResult` gets executed immediately with a `null` intent instead of waiting for the callback from the browser response.

According to [this](https://github.com/openid/AppAuth-Android/issues/299) and [this](https://github.com/openid/AppAuth-Android/issues/306) issue in the AppAuth-Android repo, the only way to fix this is to use `performAuthorizationRequest(request, completionIntent, cancelIntent)` directly instead of `startActivityForResult`.

Murmand:
Closes #29


Murmand:
Found a bug, will be fixed.

Issue #73, opened 2018-03-15
Feature request: proper support for built-in parameters


no23reason:
I want to use `prompt` parameter ([RFC](https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.3.1.2.1)) to change the auth flow.

I tried using it like this:
```js
const config: {
    issuer: ISSUER_URL,
    clientId: CLIENT_ID,
    redirectUrl: "com.example://postlogin",
    scopes: ["openid", "profile", "email", "offline_access", "phone"],
    additionalParameters: { prompt: "login" }, // trying to set the prompt
},
```

This works as expected in iOS, however in Android it throws:
```
net.openid.appauth.Preconditions.checkArgument
Preconditions.java - line 132

java.lang.IllegalArgumentException: Parameter prompt is directly supported via the authorization request builder, use the builder method instead

1 net.openid.appauth.Preconditions.checkArgument Preconditions.java:132
2 net.openid.appauth.AdditionalParamsProcessor.checkAdditionalParams AdditionalParamsProcessor.java:62
3 net.openid.appauth.AuthorizationRequest$Builder.setAdditionalParameters AuthorizationRequest.java:854
4 com.reactlibrary.RNAppAuthModule.authorizeWithConfiguration RNAppAuthModule.java:276
5 com.reactlibrary.RNAppAuthModule.access$000 RNAppAuthModule.java:41
6 com.reactlibrary.RNAppAuthModule$1.onFetchConfigurationCompleted RNAppAuthModule.java:110
7 net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute AuthorizationServiceConfiguration.java:364
8 net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute AuthorizationServiceConfiguration.java:305
9 android.os.AsyncTask.finish AsyncTask.java:667
```

[...]

PR #74, opened 2018-03-15
Prevent exception in Android when using built-in parameters

no23reason:
As described in #73, AppAuth in Android throws when some of the built-in parameters are used in the `additionalParameters` field.

As iOS version of AppAuth [does not handle those differently](https://github.com/openid/AppAuth-iOS/blob/f7b6bafd7593950e16819806f5d060a131a17228/Source/OIDTokenRequest.m#L26-L66) (and works as is), only the Android version is fixed to prevent the exceptions.

Additionally, the TypeScript declarations were updated to help users prevent typos in the enum-valued built-in parameters.


kadikraman:
Released in [v2.4.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v2.4.0) 🎉

PR #92, opened 2018-04-10
Support returning authorization code directly from lib instead of always exchanging for IdToken

thedoctor:
This PR adds conditional logic where the `authorize` method will now skip the token exchange step (where an authorization code is redeemed for an access token) and return the `code, state, redirectUri` to the JS thread directly.

TODO: iOS

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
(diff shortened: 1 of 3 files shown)


thedoctor:
Meant to PR to forked master, sorry everybody

kadikraman:
No worries 😊- look forward to the finished PR!

PR #97, opened 2018-04-17
Use correct clientSecret param in android native refresh implementation

itaym2:
Before my change, the client secret used in `refresh` was taken from `this.clientSecret`.
`this.clientSecret` was only initialized in the `authorize` method,  so unless you called `authorize` before calling `refresh`,  the refresh operation would fail.  
This PR makes the `refresh` method use the clientSecret received as an argument, instead of `this.clientSecret`


kadikraman:
^^ whoops, got a bit trigger happy there 😅 - good spot! Thanks for the PR.

kadikraman:
Released in [v2.4.1](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v2.4.1) 🚀

PR #103, opened 2018-04-25
FEATURE: Make a logout request to IdP

baltuonis:
This is a quick fix for making logout requests to identity provider endpoint.

This is really necessary since we have no other way of logging out inside browser view.

Now only works for Android. I was trying to do it for iOS, but it seemed that callback is being processed directly by AppAuth-iOS, so I was just only getting crashes.

I know this code is really just a dirty patch, however I'm really pressure by time contraints. If someone could help me refactor this and EVEN do iOS implementation would be grateful.


Example/App.js
```diff
@@ -55,10 +60,23 @@ export default class App extends Component<{}, State> {
...
-        500
+        400
       );
     } catch (error) {
-      Alert.alert('Failed to log in', error.message);
+      console.log({ error })
+
+      if (error.code === 'RNAppAuth Error') {
+        const msg = error.message;
+        if (msg === 'User cancelled flow') {
```
Aftab22: the same wont work on ios//iPhone. add "User cancelled." (including ".")  to make it work on ios//iPhone.
`if (msg === 'User cancelled flow' || msg === 'User cancelled.' )`

Aftab22:
```suggestion
        if (msg === 'User cancelled flow' || msg === 'User cancelled.') {
```

kadikraman:
Closing since this is very outdated now. Please let me know if there is still any interest in this feature and we can revisit it

PR #107, opened 2018-05-23
Fix ClientSecret creation

AtticusFetch:
Please feel free to close this one if this this issue was happening because of the server configuration, but I would really appreciate some insight.

So the problem was:

In order to make oauth - I need to pass client secret, but with `ClientSecretBasic` it was failing with error like `Each request needs to have only one auth method` or something. Without secret - I couldn't  log in.

Is this not even an issue?

Many thanks in advance!

### DIFF
--- a/android/src/main/java/com/reactlibrary/RNAppAuthModule.java
+++ b/android/src/main/java/com/reactlibrary/RNAppAuthModule.java
@@ -26,7 +26,7 @@ import net.openid.appauth.AuthorizationResponse;
 import net.openid.appauth.AuthorizationService;
 import net.openid.appauth.AuthorizationServiceConfiguration;
 import net.openid.appauth.ClientAuthentication;
-import net.openid.appauth.ClientSecretBasic;
+import net.openid.appauth.ClientSecretPost;
 import net.openid.appauth.ResponseTypeValues;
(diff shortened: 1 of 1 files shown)


kadikraman:
Hi @AtticusFetch - I won't be able to merge this, because `ClientSecretBasic` and `ClientSecretPost` do different things.

**ClientSecretBasic** - includes a client ID and client secret as an HTTP Basic Authorization header.
**ClientSecretPost** - includes a client ID and client secret as additional request parameters.

One option would obviously be to update your server to accept the client secret in the header as
`Authorization: Basic QWxhZGRpbjpPcGVuU2VzYW1l`

Or you could add the client id and secret yourself in [additional params](https://github.com/FormidableLabs/react-native-app-auth#config) (additional params just get added to the request so that might work)

kadikraman:
Add, I see! In that case, just don't pass `clientSecret` in the config in the top level, only add it to `additionalParameters`, [that will skip the `BasicAuth` step](additionalParameters).

AtticusFetch:
Tried it just now, it fails with `Failed exchange token`. I'll try to debug it a bit later and see what's wrong.

Thanks again for the advice!

Issue #96, opened 2018-04-15
Getting the authorization_code


authuser:
Hello,

Is there documentation somewhere which describes how to perform *only* the authorize request (do not perform the token request)? It would be very useful to be able to get the authorization_code.

The main use case for this is if your OAuth provider requires a non-shareable client secret, which according to the readme means that the authorization_code -> access_token exchange should be done on the server side. In order to do this with this library, we would need to be able to get the authorization_code somehow. My understanding is that this is not currently possible?

From my reading of the source, it seems reasonably straightforward to do this for android: 

1. Remove the token request calls from onActivityResult
2. Adapt tokenResponseToMap for the authorization_code to return to RN

However I am not sure how to do this for iOS. There doesn't seem like a nice way to separate the token request from the authorize request as far as I can tell. That said I am not very familiar with Objective-C, so if anyone can point me in the right direction, I would greatly appreciate it.

If none of this is at all possible, how is it recommended to do *only* the authorize request for an openId connect integration on RN? If at all possible I would like to use AppAuth, but doing the authentication/authorization flow is possible to do manually. The main challenge here is how to implement Custom Tabs (Android) and `SFAuthenticationSession/SFSafariViewController` on iOS. I think there is a react-native package for Custom Tabs, but I've yet to see anything (other than this excellent library) to implement `SFAuthenticationSession/SFSafariViewController`.

If anyone could point me in the right direction, I would greatly appreciate it.

authuser:
It appears that this has already been attempted in the GemHQ fork: https://github.com/GemHQ/react-native-app-auth/

I can't comment on how stable it is yet, but it seems to demonstrate how it could be done at least. Thanks @thedoctor and whoever else contributed!

thedoctor:
@authuser We did have basically the same requirements and, as you noted, found modifications had to be made. So I hacked in the changes from this commit: https://github.com/GemHQ/react-native-app-auth/commit/1e242edd97f00c9933173ea097f9655b09755894
I wasn't sure if the goal of this project was just to support the normal [OIDC](https://openid.net/connect/) flow for serverless authentication or if the maintainers would be interested in modifying it to have more broad OAuth facilities. 
I won't be able to find the time to make a proper PR myself for awhile, but I think if people want it, then the module's public API deserves a bit of refactoring.

addmaths07:
[...]

PR #110, opened 2018-05-30
Hacky mechanism for preventing token exchange step 

thedoctor:
This PR (discussed in https://github.com/FormidableLabs/react-native-app-auth/issues/96) adds support for a boolean `additionalParameter` called `skipTokenExchange` which will allow users to retrieve the OAuth authorization_code before it's exchanged for an access_token and id_token.

This is necessary for e.g. using this library in an application with a backend server that needs to access a user resource via client_secret (as opposed to the serverless OIDC flow).


kadikraman:
Thanks for the PR! I'm working on making this production-ready, I'll merge it when I'm ready to make a PR to change the API to the format we discussed.

Issue #112, opened 2018-05-31
Argument list too long: recursive header expansion failed


sjmorrow:
After integrating react-native-app-auth with our project, our iOS build is now failing with the following error:

```=== BUILD TARGET RNAppAuth OF PROJECT RNAppAuth WITH CONFIGURATION Debug ===


Check dependencies

Argument list too long: recursive header expansion failed at /Users/username/app_name/node_modules/react-native-app-auth/ios/../../../ios/build/Index/DataStore/v5/records/UW.
```

The problem seems to be the result of an overly aggressive HEADER_SEARCH_PATHS build setting causing Xcode to give up after it recurses through too many sub-directories for RNAppAuth's required header files. Other react native libraries have similar issues reported, and have resolved the issue by modifying HEADER_SEARCH_PATHS to be more specific than the current `$(PROJECT_DIR)/../../../ios/**`

I tried setting the HEADER_SEARCH_PATHS to `$(PROJECT_DIR)/../../../ios/Pods/**` but that produced other build errors.

Other libraries with similar problems:
https://github.com/react-native-community/react-native-camera/issues/1407
https://github.com/Microsoft/AppCenter-SDK-React-Native/issues/209

nainishmodi:
Hi..good morning @sjmorrow...have you faced @https://github.com/FormidableLabs/react-native-app-auth/issues/113  this kind of issue ?..i am stuck with this..can you help me to resolve it ? please

kadikraman:
Yes, looks like it's an error in the newest versions of RN, I've just tried it on the latest version and getting a similar error.

```
/Users/kadikraman/TestNewReact/node_modules/react-native-app-auth/ios/RNAppAuth.m:3:9: fatal error: 'AppAuth/AppAuth.h' file not found
#import <AppAuth/AppAuth.h>
```

Thanks for the tip, I'll have a look updating the HEADER_SEARCH_PATHS 🤔

kadikraman:
Yes, looks like it's an error in the newest versions of RN, I've just tried it on the latest version and getting a similar error.

Thanks for the tip, I'll have a look updating the HEADER_SEARCH_PATHS 🤔

nainishmodi:
Thanks sir for replying me..please let me know if something tackle on it..i am using android emulator and your help would be greatly appreciate..i got frustrated with this and my boss is shouting on me for this..
[...]

PR #114, opened 2018-06-03
More granular header search paths

kadikraman:
Fixes https://github.com/FormidableLabs/react-native-app-auth/issues/112


kadikraman:
Too eager!
```
"$(SRCROOT)/../../../ios/$(PROJECT_NAME)/**"
```
`$(PROJECT_NAME)` here will be `RNAppAuth`, not whatever the user's project is. Don't think there's any way of getting hold of that as a variable.

PR #131, opened 2018-06-26
[fix] podspec and making use of UIApplicationDelegate

hakonk:
I tried running `pod install` and ran into an error due to missing url for `s.homepage` :)


kadikraman:
Hey, thanks for the PR! Could you change the url to point to the GitHub repo (https://github.com/FormidableLabs/react-native-app-auth) instead? We don't have a landing page for this library in our main website, so that would make more sense.

hakonk:
Sure! 

I came across some other stuff as well that I’d like to add to the PR. Will push more code tomorrow

hakonk:
@kadikraman Here are some additional changes :)

* It seems the source code path was incorrectly set. 

* Regarding retrieving a reference to the `rootViewController` that should present the web view instance or the `SFAuthenticationViewController` instance, I've proposed calling on an `id` instance which conforms to `UIApplicationDelegate` instead. 

* Also, in order to decouple the this lib from the app, I propose we introduce a protocol `RNAppAuthAuthorizationFlowManager` for setting the flow session on the given app's instance that implements `UIApplicationDelegate`.

hakonk:
Forgot to mention how I propose on can make use of the lib in a given app:
```objective-c
@interface AppDelegate()<RNAppAuthAuthorizationFlowManager> {
  id <OIDAuthorizationFlowSession> _currentSession;
}
@end

@implementation AppDelegate
-(void)setCurrentAuthorizationFlowSession:(id<OIDAuthorizationFlowSession>)session {
  // retain session for further use
  _currentSession = session;
}
- (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<NSString *, id> *)options {
  BOOL shouldOpenUrl = [_currentSession resumeAuthorizationFlowWithURL:url];
  _currentSession = nil;
  return shouldOpenUrl;
}
@end
```

kadikraman:
@hakonk that's brilliant, it's been on my To Do list forever, but I've not had the time, really appreciate it, it's definitely the right way to go.

Could you please also update the Example app and the Readme 🙏

hakonk:
@kadikraman Yep! I'm on it now. I do have a question about the Swift section in the documentation. Is it really a use case for React Native apps where AppDelegate is implemented in Swift? While I guess it's doable, it doesn't really provide much convenience, especially since React Native for iOS is heavily based on Objective-C, Macros etc.

hakonk:
Also, I'm having some issues with building the example project. Those issues are not related to this PR however as it seems they have been there for a while. Fixing the example project, I believe should be a  separate PR. What are your thoughts?

kadikraman:
**Swift:** those instructions were added by a contributor. Figured there was no harm in merging them, although we don't officially support it. If it will no longer be valid after this PR, perhaps it's best just to delete them.

**Example app:** agreed, definitely a separate PR. It's frustrating, but I can't reproduce the error at all. The example app works for me both from Xcode and terminal on two separate computers.

hakonk:
I did actually manage to make the example project compile 😄 (see latest Podfile and xcode project changes). Regarding Swift, I could just add to the README that it's untested, but should work in a similar fashion?

hakonk:
@kadikraman I've added a suggestion in the README file concerning Swift usage.

README.md
```diff
@@ -251,91 +251,60 @@ your `Info.plist` as follows:
...
 
-You need to have a property in your AppDelegate to hold the auth session, in order to continue the
-authorization flow from the redirect. To add this, open `AppDelegate.h` in your project and add the
-following lines:
+You need to retain the auth session, in order to continue the
+authorization flow from the redirect. Follow these steps: 
 
-```diff
-+ @protocol OIDAuthorizationFlowSession;
+`RNAppAuth` will call on the given app's delegate via `[UIApplication sharedApplication].delegate`.
```
kadikraman: First step needs to be to add the imports:
```diff
+#import <AppAuth/AppAuth.h>
+#import "RNAppAuthAuthorizationFlowManager.h"
```

kadikraman:
Just tested your branch on a fresh RN project and it works! 🎉 Just needs a small update to the readme to add the imports to `AppDelegate.m`.

hakonk:
Glad to hear that! I've added some lines about imports in `AppDelegate`.

hakonk:
@kadikraman What does your release cycle look like, any scheduled releases in the near future?

hakonk:
Alright, sounds good. Let me know if you need anything from me.

PR #138, opened 2018-07-10
Bump AppAuth@0.93 for Example app

imranolas:
- Fixes the broken example app.
- Adds the missing pod files.
- Bumps AppAuth@0.93 for IOS.


kadikraman:
This builds the app, but the auth step still fails, now with
```
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[AppDelegate setCurrentAuthorizationFlow:]: unrecognized selector sent to instance 0x600000037660'
```

imranolas:
That's odd, the auth step works for me. Can you try building it with Xcode rather than the cli?

kadikraman:
Ah yes, my bad I think. It worked after a clean 👍

PR #144, opened 2018-07-12
Fix registration endpoint key in Android

NickBurkhartBB:
There is a mis-match of the key string in Android when you override the default configuration using serviceConfiguration key. If you set the registrationEndpoint it throws an exception.


kadikraman:
Thanks for your PR. Published as `v3.0.1` 🎉

PR #147, opened 2018-07-17
Android/Feature: Add method to prefetch once, reuse AutherizationServiceConfig

creambyemute:
Hey there, please have a look at this.

What it does:
Adding a method to prefetch once and reuse the AuthorizationServiceConfig with discovery mode.
It takes at least 50% less time to start up the Chrome Custom Tab on Android in Emulator if you call this in componentDidMount.
On device it's almost instant with this addition.

Usage in JS:
`import { prefetchOnce } from "react-native-app-auth";`

`componentDidMount() {
        prefetchOnce(authConfig);
}`

Edit: Refactored the original PR
Edit2: I also have a branch which adapts the example to use prefetchOnce
Edit3: Had a look at iOS, I think we don't have to do anything there. The SFSafariViewController opens instantly no matter if it has to discover or not

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,12 +41,12 @@ export const warmUpChromeCustomTab = ({
       clientId,
       scopes,
       serviceConfiguration,
-      dangerouslyAllowInsecureHttpRequests
+      dangerouslyAllowInsecureHttpRequests,
     ];
 
     RNAppAuth.warmUpChromeCustomTab(...nativeMethodArguments);
   }
-}
+};
 
 export const authorize = ({
   issuer,
--- a/index.js
+++ b/index.js
@@ -44,7 +44,7 @@ export const prefetchOnce = async ({
       dangerouslyAllowInsecureHttpRequests,
     ];
 
-    await RNAppAuth.prefetchOnce(...nativeMethodArguments);
+    RNAppAuth.prefetchOnce(...nativeMethodArguments);
   }
 };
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -22,6 +22,32 @@ const validateClientId = clientId =>
...
+
+    const nativeMethodArguments = [
+      issuer,
+      redirectUrl,
+      clientId,
+      scopes,
+      serviceConfiguration,
+      dangerouslyAllowInsecureHttpRequests,
+    ];
+
```
creambyemute: Idk, I'm relatively new to JS and promises, but we should return/await here but it gives linting error. Is it maybe because of the eslint version?

I'd be happy to change it if I could just find out to what... :(

kadikraman: No need do `await` here, because you're returning a function that is `await`ed when it's called

kadikraman:
Thanks for the PR 🙏 

Seems to work fine for me, and it does significantly speed up the Chrome Custom Tab, I think it's a really good idea.

The method name could be a bit clearer, how about `prefetchConfiguration`? Just to make it a bit more obvious what it's doing.

Also would you be able to update the readme?

creambyemute:
Yeah good points! I'll edit the PR tonight or tomorrow.

tamcgoey:
@creambyemute just curious if this is moving forward, would love to get the added speed update.

creambyemute:
@tamcgoey @kadikraman We've unfortunately moved away from app-auth due to the requirement for always available offline login in our App, so that's why I haven't done the required adaptions yet.

I'll try to quickly do this this weekend.

michaelrhughes:
I've created a new pull request which I've rebased this one to the latest master and addressed the code review comments:
https://github.com/FormidableLabs/react-native-app-auth/pull/289

kadikraman:
Closing this, as it has been merged a released in https://github.com/FormidableLabs/react-native-app-auth/pull/289

PR #168, opened 2018-08-31
remove scope field for token exchange

### LINKED ISSUE #140: Remove parameter during token exchange?
Is it possible to remove scope parameters during token exchange?
I am trying to implement Salesforce OAuth2 with refresh_token flow, and I traced back into "AuthorizationService.java" which put the "scope" parameter in the body.

Currently Salesforce response `{"error":"invalid_request","error_description":"scope parameter not supported"}` because it writes scope parameter onto the body. Ideal implementation would be like https://github.com/golang/oauth2/pull/146/files#diff-6fd6bf47c9136acf10e9329aca75387aL104

version:`"react-native-app-auth": "2.6.0"`

mikaoelitiana:
provides a quick solution to remove scope field when requesting token
fixes #140

### DIFF
--- a/android/src/main/java/com/reactlibrary/RNAppAuthModule.java
+++ b/android/src/main/java/com/reactlibrary/RNAppAuthModule.java
@@ -40,6 +40,7 @@ import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
+import java.lang.reflect.Field;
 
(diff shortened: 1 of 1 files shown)


kadikraman:
Hi, thanks a lot for the PR!

I agree with that @SuhanKoh said in the issue you referenced in that we can't just remove the scope, because though it may fix your issue, it will not work for everyone.

So I think we'll have to go the long way around and separate the token exchange step as originally planned.

PR #169, opened 2018-09-03
Support making nonce nullable

FelipeBuiles:
AppAuth-iOS worked correctly with [AWS Cognito](https://aws.amazon.com/cognito/) up until its 0.92 release, later they added support that added support for a nonce parameter. The issue lies in the inabilty of Cognito to handle the nonce parameter (mentioned in their forums [here](https://forums.aws.amazon.com/thread.jspa?messageID=802423&tstart=0)).
There is a comment on the `OIDAuthorizationRequest.h` file from AppAuth-iOS that mentions something about it being nullable:
```
    @param nonce String value used to associate a Client session with an ID Token. Can be set to nil
        if not using OpenID Connect, although pure OAuth servers should ignore params they don't
        understand anyway.
```

The idea here is to add a new optional param `useNonce` that when set to false, tells the iOS implementation to not generate a nonce attribute automatically. This only happens on the iOS implementation, Android does not have this issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,6 +28,7 @@ export const authorize = ({
   clientId,
   clientSecret,
   scopes,
+  useNonce = true,
   additionalParameters,
   serviceConfiguration,
   dangerouslyAllowInsecureHttpRequests = false,
@@ -48,6 +49,10 @@ export const authorize = ({
   ];
   if (Platform.OS === 'android') {
     nativeMethodArguments.push(dangerouslyAllowInsecureHttpRequests);
+  } else {
+    // add a new useNonce param on iOS to support making it optional
+    const nonceParamIndex = 5;
+    nativeMethodArguments.splice(nonceParamIndex, 0, useNonce);
   }
 
   return RNAppAuth.authorize(...nativeMethodArguments);
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -362,7 +362,7 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
             @Override
             public void onTokenRequestCompleted(@Nullable TokenResponse response, @Nullable AuthorizationException ex) {
                 if (response != null) {
-                    WritableMap map = tokenResponseToMap(response, null);
+                    WritableMap map = tokenResponseToMap(response);
                     promise.resolve(map);
                 } else {
                     promise.reject("RNAppAuth Error", "Failed refresh token");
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -48,6 +49,10 @@ export const authorize = ({
   ];
   if (Platform.OS === 'android') {
     nativeMethodArguments.push(dangerouslyAllowInsecureHttpRequests);
+  } else {
+    // add a new useNonce param on iOS to support making it optional
+    const nonceParamIndex = 5;
```
kadikraman: This seems a bit fiddly, I'd prefer to just add it to the end as the android-only method above
```js
nativeMethodArguments.push(useNonce);
```

kadikraman:
Hi @FelipeBuiles thanks a lot for the PR!

Since this is AWS Cognito specific, would you be able to add a config example (something like [this](https://github.com/FormidableLabs/react-native-app-auth#identity-server-3)).

Could you also please add the new parameter to the [config](https://github.com/FormidableLabs/react-native-app-auth#config).

kadikraman:
Hi @FelipeBuiles @benjarwar I was planning on merging your contribution, but I am now confused about why it's necessary.

I've added documentation for [aws-cognito](https://github.com/FormidableLabs/react-native-app-auth#aws-cognito) in our readme, but I cannot reproduce this issue at all and can authenticate with using AppAuth v0.94.

Here's the config for my example app:

```
{
  clientId: '717b63pm9mt5btbhdhmt89qikf',
  redirectUrl: 'com.myclientapp://myclient/redirect',
  serviceConfiguration: {
    authorizationEndpoint: 'https://app-auth-test.auth.eu-west-1.amazoncognito.com/oauth2/authorize',
    tokenEndpoint: 'https://app-auth-test.auth.eu-west-1.amazoncognito.com/oauth2/token',
    revocationEndpoint: 'https://app-auth-test.auth.eu-west-1.amazoncognito.com/oauth2/revoke'  }
}
```

dreyks:
actually i have this same issue with Dropbox API so I have high hopes for seeing that merged

dreyks:
ugh... tried @FelipeBuiles fork, now getting `unknown field 'code_challenge'`. FFS dropbox!

So this PR alone won't help me. Also need to turn off PKCE as described in https://github.com/openid/AppAuth-iOS/issues/305#issuecomment-424472617

dreyks:
nvm. Dropbox refuses to redirect to urls not starting from `https` in responseType `code`. And `token` is not supported by AppAuth so I have to find another OAuth library

nol13:
same issue moving from 2.4.1, so far so good with this fix.

think we're running our own identity server instance and not cognito, which is mostly out of my domain, so possibly config update there would do it too, but this is working for us.

kadikraman:

PR #232, opened 2019-01-31
Add `usePKCE` Flag for Dropbox

awmichel:
This allows providers like Dropbox who don't support PKCE and go so far as to error is you provide unknown parameters in the OAuth2 request to function correctly.

By specifying `usePKCE: false` in the auth options the `code_challenge` parameter will not be sent to the OAuth2 provider and code verification will not run against the response.

Related to #231 and #169. See comment https://github.com/FormidableLabs/react-native-app-auth/pull/169#issuecomment-431587776 as well.

The redirect URL issue with Dropbox can be resolved using universal links or a redirect service.

### DIFF
--- a/ios/RNAppAuth.m
+++ b/ios/RNAppAuth.m
@@ -39,6 +39,7 @@ RCT_REMAP_METHOD(authorize,
                  additionalParameters: (NSDictionary *_Nullable) additionalParameters
                  serviceConfiguration: (NSDictionary *_Nullable) serviceConfiguration
                  useNonce: (BOOL *) useNonce
+                 usePKCE: (BOOL *) usePKCE
                  resolve: (RCTPromiseResolveBlock) resolve
                  reject: (RCTPromiseRejectBlock)  reject)
 {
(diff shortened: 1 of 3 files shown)


ios/RNAppAuth.m
```diff
@@ -205,7 +209,7 @@ - (void)authorizeWithConfiguration: (OIDServiceConfiguration *) configuration
                                                        strongSelf->_currentSession = nil;
                                                        if (authState) {
                                                            resolve([self formatResponse:authState.lastTokenResponse
-                                                               withAdditionalParameters:authState.lastAuthorizationResponse.additionalParameters]);
+                                                               withAdditionalParameters:authState.lastTokenResponse.additionalParameters]);
```
awmichel: I could use some additional thoughts and opinions on this change. What is the reasoning behind returning the `additionalParameters` from the authorization response instead of the token response? Dropbox returns the UID in the token response, so this change allowed me to pull that out. Otherwise there is no way to access it.

awmichel:
@kadikraman Thanks for taking a look. I resolved the merge commits and separated the `additionalParameters` for the authorize and token responses.

It looks like the Android library was also merging the additional parameters from the `authorize` call into the `additionalParameters` it returned. I went ahead and removed that code so iOS and Android are now consistent.

Let me know if there are any other tweaks you'd like. There are a couple breaking changes in here now to signal out.

I also added a section to the README for configuring Dropbox.

kadikraman:

PR #242, opened 2019-02-07
Replace 'compile' with 'implementation' in android/build.gradle

### LINKED ISSUE #241: WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation'.
## Issue

Current build throws this warning on build.gradle for the use of 'compile' instead of 'implementation' for dependencies. This was deprecated last year and is now obsolete.

---

## Environment

- **Your Identity Provider**: `e.g. IdentityServer 4 / Okta / Azure`
- **Platform that you're experiencing the issue on**:
  - [ ] iOS
  - [x] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [ ] Both
- **Are you using Expo, e.g. `ExpoKit`?**
  - [x] No
  - [ ] Yes, I've _not_ ejected
  - [ ] Yes, but I **have** ejected to `ExpoKit`
  - [ ] Yes, but I **have** ejected to vanilla React Native

tsmith5:
- Fixes #241 
- Replaces the now obsolete 'compile' for build.gradle dependencies with the newer 'implementation'
- Compiling any app that depends on this package should no longer complain about the use of 'compile'
- No updates to tests or readme required, just cleans up a warning.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -34,6 +34,6 @@ repositories {
 }
 
 dependencies {
-    compile 'com.facebook.react:react-native:+'
-    compile "net.openid:appauth:0.7.1"
+    implementation 'com.facebook.react:react-native:+'
+    implementation "net.openid:appauth:0.7.1"
 }


sraka1:
It has to be noted that this will break the build for anyone that's using RN < 0.57.

So this would a breaking change and version bump...

kadikraman:
Good point @sraka1 - I'll add a note to the Readme. There will be a couple of breaking changes in the next release, so it'll definitely be a major version bump and this can be included.

tsmith5:
@sraka1 Sorry to be a bother, but could you explain how this change breaks RN < 0.57? I wasn't aware of it until you mentioned it. Thanks!

sraka1:
@tsmith5 until 0.57 RN used Gradle2, which lacks the `implementation` directive introduced in Gradle3. Furthermore, while Gradle3 emits a warning about this (deprecated, not removed), the build would still pass meanwhile on Gradle2 (so RN < 0.57) the build will definitely fail...

sraka1:
Personally, I’ll have to monkey-patch (or fork this package) to maintain compatibility with Gradle2 as we’re not ready to upgrade RN and/or Gradle just yet...

kadikraman:
@sraka1 that's actually a good point. Perhaps we should revert this and add a suggestion for using [patch-package](https://github.com/ds300/patch-package) instead.

tsmith5:
Just a minor clarification for anyone else reading this in the future:

This doesn't break the React Native package, what it breaks is Gradle <3. If you are running an older version of React Native (eg, 0.55.4) and upgrade to Gradle3 it will build fine.

barrymichaeldoyle:
The README.md sends me here but doesn't really explain how to use patch-package to solve the compile/implementation problems.

Are there any guides out there regarding how to use patch-package in this context?

kadikraman:
I see it has been a common misconception, so I'm sorry for the confusion. If you don't mind the warnings, feel free to **do nothing**.

If you are using `react-native-app-auth@5.0.0` or above **DO NOTHING** - this is already included in the release.

If however you do care about the warnings and you're on Gradle 3, you can do the following:

1. open `node_modules/react-native-app-auth/android/build.gradle` and mirror the changes in this PR
2. run `npx patch-package some-package` (this will create a `.patch` file)
3. commit the patch to your source control
4. in `package.json` add a postistall script:
```diff
 "scripts": {
+  "postinstall": "patch-package"
 }
```
5. add the library as a dependency
```
yarn add patch-package postinstall-postinstall
```
or
```
npm i patch-package
```
This basically adds a patch to the library every time you do an install. Once you can upgrade to v5 of this library you can simply delete the patch file and the package.

williamtorres1:
I'm still getting this error on RN0.61.1 and RN0.64.0. The steps on the comment above solved the problem.

inklesspen1rus:
Maybe it's time to merge that?
There is already RN0.71.11
World has Gradle 8

PR #246, opened 2019-02-11
Explicitly add invariant to peerDependencies

pettyalex:
invariant should be listed as a peer dependency as well, rather than an assumption that react-native will always have a version of invariant that this module is able to use.

I can create an issue to match this PR if the maintainers would prefer, but it's purely an update to the package.json dependencies that makes it clear where invariant is coming from in production.

- [ ] include issue number that will be resolved by this PR (e.g. `Fixes #1234`)
- [ ] include a summary of the work
- [ ] include steps to verify
- [ ] update tests (if applicable)
- [ ] update readme (if applicable)
- [ ] update typescript definitions (if applicable)

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "author": "kadi.kraman@formidable.com",
   "license": "MIT",
   "peerDependencies": {
+    "invariant": "^2.2.2",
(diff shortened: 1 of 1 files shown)


kadikraman:

PR #259, opened 2019-02-27
Fix null response for 'scope'

zzorba:
If the authResponse does not include a 'scope' parameter, the app crashes on android as it parses the response.  This restores the behavior that existed before 4.0, where a missing scope would be just fine.  Since you might not have control over the server, it would be nice to make this more tolerant to missing fields (like it is on IOS)


kadikraman:
Published as a patch [v4.1.1](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v4.1.1) 🎉

Issue #271, opened 2019-03-14
[Android] Add flag for Client Authentication Method


kpelelis:
There are some cases in which a client needs to select the strategy in which the oauth credentials should be sent to the server.  AppAuth for Android uses [two classes](https://github.com/openid/AppAuth-Android#utilizing-client-secrets-dangerous) for that purpose. The Android side of the React Native library seems to be using [basic authentication](https://github.com/FormidableLabs/react-native-app-auth/blob/master/android/src/main/java/com/rnappauth/RNAppAuthModule.java#L245) by default.

PR #272, opened 2019-03-14
[Help Needed] Configure client authentication strategy

kpelelis:
- [X] include issue number that will be resolved by this PR (e.g. `Fixes #1234`)
- [X] include a summary of the work
- [X] include steps to verify
- [X] update tests (if applicable)
- [X] update readme (if applicable)
- [X] update typescript definitions (if applicable)

Closes: #271

## Summary

This patch introduces a `clientAuthMethod` flag in both `authroize` and `refresh` methods that utilizes both ClientSecretBasic, a class used for secret exchange with basic authentication (default), and ClientSecretPost, which send authentication credentials via the HTTP POST body.

This PR does not break the default behavior. 

**This feature is currently for Android as I do not have Mac to verify that this is indeed a case for apple devices.**

## Steps to verify

1. Verify that the examples are still working as expected, since this patch will not break the existing functionality.
 2. Set the `clientAuthMethod` to `post` in the configuration object and verify (using a packet sniffer or the logs of the server) that the credentials are exchanged using the HTTP body.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ with optional overrides.
 * **additionalParameters** - (`object`) additional parameters that will be passed in the authorization request.
   Must be string values! E.g. setting `additionalParameters: { hello: 'world', foo: 'bar' }` would add
   `hello=world&foo=bar` to the authorization request.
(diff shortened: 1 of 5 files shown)


kadikraman:
Released in [v4.3.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v4.3.0)

Issue #279, opened 2019-03-25
Disable PKCE on Android


sebk:
In iOS it is possible to configure PKCE with the 'usePKCE' config flag.

Is it possible to do the same on Android?
When it is not possible, are you planning to integrate this configuration (like in iOS)?

**Edit:**
I found a way to disable it on Android. I could create a pull request.

PR #284, opened 2019-03-29
Use pkce android

sebk:
- [x] Implements #279
- [x] Summary of the work: Moved the parameter `usePKCE` to the general config (for Android and iOS), to be able to disable PKCE also on Android.
- [ ] include steps to verify
- [x] update tests (if applicable)
- [x] update readme (if applicable)
- [ ] update typescript definitions (if applicable)

### DIFF
--- a/index.js
+++ b/index.js
@@ -76,6 +76,7 @@ export const authorize = ({
     scopes,
     additionalParameters,
     serviceConfiguration,
+    usePKCE
   ];
 
   if (Platform.OS === 'android') {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -86,7 +87,6 @@ export const authorize = ({
 
   if (Platform.OS === 'ios') {
     nativeMethodArguments.push(useNonce);
-    nativeMethodArguments.push(usePKCE);
```
kadikraman: I know this is not ideal, but could you add the `usePKCE` argument to Android without disrupting the parameter order on iOS? Otherwise this will be a breaking change, and I won't be able to release it until the next major version.

sebk: I pushed a new commit. I hope I did it correctly. :)

kadikraman: Perfect, thanks!

kadikraman:
Released in [v4.3.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v4.3.0)

PR #289, opened 2019-04-09
Adds a method to prefetch AuthorizationServiceConfiguration

### LINKED ISSUE #145: Android: Any way of speeding up opening of chrome custom tab?
On android it takes quite some time (up to 2 seconds) to open up the chrome custom tab after clicking on the button.

I first thought I did something wrong in my app but then tried the example app on android and it's the same there.

On iOS it almost instantly opens up the SFSafariViewController.

I know chrome custom tabs can be "warmed up" with the url to launch and I see that in AppAuth-Android there also is a code part which does that, but it seems as if it happens too late.

Edit: I had a look at the implementation and in AppAuth-Android and in react-native-app-auth and was able to speed up the opening of chrome custom tabs quite a bit with a new method "warmUpChromeCustomTabs" which I call in `componentDidMount` in my project.

I'll create a PR for it but it'll probably need some improvement/refactoring as it was just a quick test to see if it works.

See: https://github.com/FormidableLabs/react-native-app-auth/pull/147

Cheers :-)

michaelrhughes:
This PR is a rebase of https://github.com/FormidableLabs/react-native-app-auth/pull/147.

- I have also addressed the code review comments so this PR should be good to go.
- I have added in warming up and prefetching for Chrome Custom Tabs

Issue Resolved:
- Fixes #145

Summary of the Work:
- Change exposes a method that prefetches the AuthorizationServiceConfiguration and warms the chrome custom tab

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,12 +81,13 @@ const result = await authorize(config);
 
 ### `prefetchConfiguration`
 
-This will prefetch the authorization service configuration. Invoking this function is optional
-and will speed up calls to authorize.
+ANDROID This will prefetch the authorization service configuration. Invoking this function is optional
+and will speed up calls to authorize. This is only supported on Android.
 
 ```js
 import { prefetchConfiguration } from 'react-native-app-auth';
 
+const warmAndPrefetchChrome = true
 const config = {
   issuer: '<YOUR_ISSUER_URL>',
   clientId: '<YOUR_CLIENT_ID>',
@@ -94,7 +95,7 @@ const config = {
   scopes: ['<YOUR_SCOPES_ARRAY>'],
 };
 
-prefetchConfiguration(config);
+prefetchConfiguration(warmAndPrefetchChrome, config);
 ```
 
 #### config
--- a/README.md
+++ b/README.md
@@ -79,6 +79,24 @@ const config = {
 const result = await authorize(config);
 ```
 
+### `prefetchConfiguration`
+
+This will prefetch the authorization service configuration. Invoking this function is optional
+and will speed up calls to authorize.
+
+```js
+import { prefetchConfiguration } from 'react-native-app-auth';
+
+const config = {
+  issuer: '<YOUR_ISSUER_URL>',
+  clientId: '<YOUR_CLIENT_ID>',
+  redirectUrl: '<YOUR_REDIRECT_URL>',
+  scopes: ['<YOUR_SCOPES_ARRAY>'],
+};
+
+prefetchConfiguration(config);
+```
+
 #### config
 
 This is your configuration object for the client. The config is passed into each of the methods
--- a/README.md
+++ b/README.md
@@ -87,15 +87,15 @@ and will speed up calls to authorize. This is only supported on Android.
 ```js
 import { prefetchConfiguration } from 'react-native-app-auth';
 
-const warmAndPrefetchChrome = true
 const config = {
+  warmAndPrefetchChrome: true,
   issuer: '<YOUR_ISSUER_URL>',
   clientId: '<YOUR_CLIENT_ID>',
   redirectUrl: '<YOUR_REDIRECT_URL>',
   scopes: ['<YOUR_SCOPES_ARRAY>'],
 };
 
-prefetchConfiguration(warmAndPrefetchChrome, config);
+prefetchConfiguration(config);
 ```
 
 #### config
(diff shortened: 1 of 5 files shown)


michaelrhughes:
@kadikraman @creambyemute FYI since you were on the previous pull request :)

README.md
```diff
@@ -79,6 +79,24 @@ const config = {
 const result = await authorize(config);
 ```
 
+### `prefetchConfiguration`
+
```
kadikraman: Could you add a note here to say this is Android only? (similar to how we do it in config docs)

android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -498,6 +574,21 @@ private AuthorizationServiceConfiguration createAuthorizationServiceConfiguratio
         );
     }
 
+    private void warmChromeCustomTab(Context context, final String issuer) {
```
kadikraman: This is a great idea, but I think this should be optional, so the user would be able to opt out of this and _only_ prefetch the configuration.

michaelrhughes: Agreed, I'll add a boolean to the prefetch method.

README.md
michaelrhughes: Yup!

michaelrhughes:
@kadikraman I've updated the PR, let me know what you think of the changes :)

kadikraman:
@michaelrhughes thanks! Would you mind adding the `warmAndPrefetchChrome` in the config object you pass to the `prefetchConfiguration` as an argument instead? I can understand wanting to separate these, but `prefetchConfiguration(true, config)` is not a very obvious api for the user. I think it would be better if all the options were in the dingle config for this method.

michaelrhughes:
@kadikraman Sure, I can make that change. My original thought process was being able to reuse the same config between `authorize` calls and `prefetch`.

michaelrhughes:
@kadikraman Do you know what eslint is complaining about here?

```
yarn run v1.9.4
$ eslint .

/workspace/repo/index.js
  0:0  error  Delete `··`  prettier/prettier
  0:0  error  Delete `··`  prettier/prettier
  0:0  error  Delete `··`  prettier/prettier
  0:0  error  Delete `··`  prettier/prettier
  0:0  error  Delete `··`  prettier/prettier
  0:0  error  Delete `··`  prettier/prettier
  0:0  error  Delete `··`  prettier/prettier
  0:0  error  Delete `··`  prettier/prettier

✖ 8 problems (8 errors, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

michaelrhughes:
@kadikraman let me know what you think, everything should be good to go :)

michaelrhughes:
Hey @kadikraman, have you gotten a chance to take a look at this pr yet? :)

michaelrhughes:
@kadikraman Have you gotten a chance to take a look at this PR yet?

kadikraman:
Published in [v4.4.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v4.4.0) - thanks a lot! 🙌

michaelrhughes:
No problem, thanks for merging it!

csath:
@michaelrhughes Nice work on the PR! 

But I have a problem.

I see you are keeping `AuthorizationServiceConfiguration` reference and and try to use already fetched configurations from there(if available) without calling OpenID well-known id endpoint.

Although `prefetchConfiguration` method is **called or not** currently fetched configurations are saved with the following code line  inside `authorize` method  `mServiceConfiguration.set(fetchedConfiguration);`

So my problem is if user try to change auth configs on the fly(from react native js code) and try to call `authorize` method it'll not use new configs instead old configs saved in mServiceConfiguration.

So this breaks the flexibility of using different service issuers in one application.
I think we need a proper way to store prefetched configs(may be a hash map)

Sample js side code to produce the problem I'm describing above. 
```js
import { authorize } from 'react-native-app-auth';

// First authorize call with configs 1
await authorize({
        issuer: PILOT_ISSUER_1,
        clientId: PILOT_CLIENT_ID_1,
        redirectUrl: REDIRECT_URL_1,
        scopes: [ 'offline_access'],
});

// Second authorize call with configs 1
await authorize({
        issuer: PILOT_ISSUER_2,
        clientId: PILOT_CLIENT_ID_2,
        redirectUrl: REDIRECT_URL_2,
        scopes: [ 'offline_access'],
});
```

Issue #130, opened 2018-06-25
Crash on Android : dataIntent must not be null


shinekumar:
Thanks for the great component. It is working perfectly on iOS. But on Android, I am getting a crash.
 I have given details below. Please let me know if you need additional details. Thanks!

**Device Info**
Samsung Galaxy S5
OS:  6.0.1

**Package Info**
"react-native-app-auth": "^2.6.0",

**Issue**

- Login Page Loads
- Enters Username and Password. Press Login
- App Crashes

**Log**

java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=0, result=0, data=null} to activity {com.realsavvyv3/com.realsavvyv3.MainActivity-Main}: java.lang.NullPointerException: dataIntent must not be null
                                                       at android.app.ActivityThread.deliverResults(ActivityThread.java:4920)
                                                       at android.app.ActivityThread.handleSendResult(ActivityThread.java:4963)
                                                       at android.app.ActivityThread.access$1600(ActivityThread.java:221)
                                                       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1848)
                                                       at android.os.Handler.dispatchMessage(Handler.java:102)
                                                       at android.os.Looper.loop(Looper.java:158)
                                                       at android.app.ActivityThread.main(ActivityThread.java:7224)
                                                       at java.lang.reflect.Method.invoke(Native Method)
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
                                                    Caused by: java.lang.NullPointerException: dataIntent must not be null
                                                       at net.openid.appauth.Preconditions.checkNotNull(Preconditions.java:55)
                                                       at net.openid.appauth.AuthorizationResponse.fromIntent(AuthorizationResponse.java:565)
                                                       at com.reactlibrary.RNAppAuthModule.onActivityResult(RNAppAuthModule.java:213)
                                                       at com.facebook.react.bridge.ReactContext.onActivityResult(ReactContext.java:255)
                                                       at com.facebook.react.ReactInstanceManager.onActivityResult(ReactInstanceManager.java:730)
                                                       at com.facebook.react.ReactActivityDelegate.onActivityResult(ReactActivityDelegate.java:126)
                                                       at com.facebook.react.ReactActivity.onActivityResult(ReactActivity.java:77)
[...]

PR #352, opened 2019-07-11
Stop onActivityResult from proceeding when intent is null to prevent app from crashing

AlgirdasVZ:
Partially fixes https://github.com/FormidableLabs/react-native-app-auth/issues/130 

## Description

There are at least several cases where the app crashes in onActivityResult method when received data intent is null (more about this - https://github.com/FormidableLabs/react-native-app-auth/issues/130). The cause of the crash is that AuthorizationResponse.fromIntent method (imported from net.openid.appauth) has a null checker method, which throws NullPointerException and the user is presented with a native crash info screen. I'm proposing to add a check for data intent being null, and terminate the authorization process if it is (reject the the promise with error message and return from onActivityResult method). As the data intent being null will always end up in crashing the app, there is no point to continue onActivityResult method. Terminating the authorization process instead would be much more user-friendly solution.

## Steps to verify

One of the scenarios to reproduce the crash:

```
1. Open the app
2. Open the idp web login screen in the app
3. Put the app in background
4. Start the app from the device launcher

Prerequisites: android:launchMode="singleTask" in AndroidManifest.xml
```

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -300,6 +300,11 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
     @Override
     public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
         if (requestCode == 0) {
+            if (data == null) {
+                promise.reject("Failed to authenticate", "Data intent is null" );
+                return;
+            }
+            
             final AuthorizationResponse response = AuthorizationResponse.fromIntent(data);
             AuthorizationException exception = AuthorizationException.fromIntent(data);
             if (exception != null) {


mklekowski:
Can it be released as a patch to 4.4.0?

kadikraman:
Hi @mklekowski - I'd like to avoid maintaining two branches of the library - it'll get messy quick!

I'd use [`patch-package`](https://github.com/ds300/patch-package) for this, since the change is so small:

1. open `node_modules/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java` and mirror the changes in this PR
2. run `npx patch-package some-package` (this will create a `.patch` file)
3. commit the patch to your source control
4. in `package.json` add a postistall script:
```diff
 "scripts": {
+  "postinstall": "patch-package"
 }
```
5. add the library as a dependency
```
yarn add patch-package postinstall-postinstall
```
or
```
npm i patch-package
```
This basically adds a patch to the library every time you do an install. Once you can upgrade to v5 of this library you can simply delete the patch file and the package.

kadikraman:

Issue #363, opened 2019-07-31
v5.0.0-rc1 does not autolink


sibelius:
## Issue

Using this commit works well with autolinking
"react-native-app-auth": "git+https://git@github.com/FormidableLabs/react-native-app-auth.git#f16014431f5c7470a8404b9f1e26bfd08ea6e72c"

But using 

 "react-native-app-auth": "v5.0.0-rc1",

does not

![image](https://user-images.githubusercontent.com/2005841/62243304-69611a00-b3b3-11e9-86f7-83beec21d632.png)

there is no `react-native-app-auth.podspec` in our node_modules, see image above

---

## Environment

* **Your Identity Provider**: v5.0.0-rc1
* **Platform that you're experiencing the issue on**: iOS
* **Are you using Expo?** No

AlgirdasVZ:
I can confirm this as well. Looks like podspec file was filtered out during package publishing. Is this maybe intended as it's still rc?

hazone:
I guess the react-native-app-auth.podspec need to be added to the files section in package.json. https://github.com/FormidableLabs/react-native-app-auth/blob/v5.0.0-rc1/package.json

danfri86:
@hazone That's correct.
I used the package "[Patch package](https://www.npmjs.com/package/patch-package)" that I already had installed for some other package fixes.
I downloaded the .podspec file and added to the root of react-native-app-auth in my node_modules folder. I also added react-native-app-auth.podspec to the files section in package.json and ran the patch package.
After this I did pod install from the ios folder and it worked.

AlgirdasVZ:
@danfri86 the downside of your approach is that it doesn't fit in  in the build automatisation process. The easiest way to solve this issue would be publishing new npm package (next rc), with podscpec file put in place, as after testing rc-1 it looks like it works quite well otherwise.

kadikraman:
Released in [v5.0.0-rc2](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v5.0.0-rc2) 🌟

PR #374, opened 2019-08-30
Add podspec to included `files`

jeffreyffs:
Fixes https://github.com/FormidableLabs/react-native-app-auth/issues/363

## Description

Add the podspec to the bundled files

## Steps to verify

Try to auto-link this library.

You should *not* see
`[!] use_native_modules! skipped the react-native dependency 'react-native-app-auth'. No podspec file was found.` (which is what currently happens with 5.0.0-rc1)

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "ios",
     "android",
     "index.js",
-    "index.d.ts"
+    "index.d.ts",
+    "react-native-app-auth.podspec"
   ],
   "repository": {
     "type": "git",


bachhuberdesign:
Any idea of when the auto-link fixes will be merged/released? It seems too critical of an issue to leave unaddressed.

jeffreyffs:
That is up to @kadikraman  :)

This only affects 5.0.0 RC 1 as far as I can tell (Which you probably shouldn't be using for production code yet).

Using manual linking should work fine until 5.0.0 is released.

kadikraman:

PR #378, opened 2019-09-09
Use the same error message code across platforms

### LINKED ISSUE #377: Refresh token error codes are not unified across platforms and providers
## Issue
When the refresh token expires (could be due to changing password, not being used for a while etc) the error message sent from the library is different depending on platform and provider. It would be very nice if this could be unified. 

### Example
Responses for Google and Azure, havn't tried the others in the table below. 

```
catch(e) {
    e.code // ErrorCode in example below
    e.message // ErrorMessage in example below
}
```

<table>
<tr><td></td><td>Android</td><td>iOS</td></tr>
<tr><td>Azure</td><td>
<b>ErrorCode:</b> Failed to refresh token
 <br>
<b>ErrorMessage:</b> AADSTS50173: The provided grant has expired due to it being revoked ....
</td><td>
<b>ErrorCode:</b> RNAppAuth Error
<br>
<b>ErrorMessage:</b> invalid_grant: AADSTS50173: The provided grant has expired due to it being revoked
</td></tr>
<tr><td>Google</td><td>
<b>ErrorCode:</b> Failed to refresh token
<b>ErrorMessage:</b> Token has been expired or revoked.
</td><td>
<b>ErrorCode:</b> RNAppAuth Error
<b>ErrorMessage:</b> invalid_grant: Token has been expired or revoked.

</td></tr>
</table>

## Suggested fix
The same error should return the same error code defined by the library. The error messages could be kept as they are.

kadikraman:
Fixes #377

## Description
The error codes are not the same across platforms, e.g. when the token refresh fails, Android error code is `Failed to refresh token` and iOS is `RNAppAuth Error`. This ensures that the error codes are the same across platforms.

## Error codes
The following error messages can be thrown:
- `service_configuration_fetch_error` - could not fetch the service configuration
- `authentication_failed` - user authentication failed
- `token_refresh_failed` - could not exchange the refresh token for a new JWT

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -268,7 +268,7 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
                                 @Nullable AuthorizationServiceConfiguration fetchedConfiguration,
                                 @Nullable AuthorizationException ex) {
                             if (ex != null) {
-                                promise.reject("token_refresh_failed", getErrorMessage(ex));
+                                promise.reject("service_configuration_fetch_error", getErrorMessage(ex));
                                 return;
                             }
 
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -96,7 +96,7 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
                     isPrefetched = true;
                     fetchConfigurationLatch.countDown();
                 } catch (Exception e) {
-                    promise.reject("RNAppAuth Error", "Failed to convert serviceConfiguration", e);
+                    promise.reject("configuration_error", "Failed to convert serviceConfiguration", e);
                 }
             } else if (mServiceConfiguration.get() == null) {
                 final Uri issuerUri = Uri.parse(issuer);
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -268,7 +268,7 @@ public void onFetchConfigurationCompleted(
                                 @Nullable AuthorizationServiceConfiguration fetchedConfiguration,
                                 @Nullable AuthorizationException ex) {
                             if (ex != null) {
-                                promise.reject("Failed to fetch configuration", getErrorMessage(ex));
+                                promise.reject("token_refresh_failed", getErrorMessage(ex));
                                 return;
```
henninghall: Should be `service_configuration_fetch_error` right?

kristerkari:
@kadikraman @henninghall Is there something pending still with this one? This would be a nice improvement to get released, as the errors are difficult to handle when there are platform differences.

kristerkari:
I'll give an example of what I'm doing and why I need these improvements to the error messages:

Currently I have a problem that sometimes refreshing the token errors out because of network issues or the connection timing out. In these cases it makes sense to just retry the refresh as the token can be still refreshed. 

With these changes I could do something like this for both Android and iOS (the message might have some differences on Android/iOS) :

```js
// an example, not working or production ready code.
.catch(err => {
  if (err.code === "token_refresh_failed" && err.message === "something") {
    refresh(AUTH_CONFIG, { refreshToken })
  }
})
```

henninghall:
@kristerkari It's all good from my side.

kadikraman:

PR #381, opened 2019-09-14
fix gradle setup

dulmandakh:
## Description

This PR removes Android Gradle Plugin use in **android** folder or RN module, so that it won't create conflicts if version mismatches. Then creates a root gradle project, build.gradle.kts at root folder, so we can use it with Android Studio.

Also removed buildToolsVersion because it's been deprecated a long ago.

## Steps to verify

Still can be used as RN module as before, but to develop you need to open not **android** folder but root folder with Android Studio.


dulmandakh:
@kadikraman please review and merge

friederbluemle:
Hi @dulmandakh - With the Android Gradle plugin removed, we're no longer able to open the `android` folder in Android Studio, or build the project individually. Did you find a way around that?

friederbluemle:
Okay sorry - I didn't see you moved the Gradle files into the root directory. Since Gradle files are only Android related, wouldn't it be better to keep them contained in the `android` folder? Just like everything iOS is in the `ios` folder. For all React Native project, I normally do `xed ios && studio android`. Now with the files moved, this is no longer possible.

I'm also still trying to understand what exactly the problem is with a nested Android Gradle plugin - It should not be used if it's not declared in the root of the project (i.e. included as a module by other projects).

dulmandakh:
In the past, I had many conflicts by modules that trying to load different versions of Android Gradle Plugin, thus causing build failures. This PR tries to resolve the issue moving AGP outside of the module.

I too want to contain everything Android in android folder, but I found this. I would be more than happy if you bring alternate solution to table.

friederbluemle:
@dulmandakh - Yes, makes total sense. I actually found an alternate solution that results in the same outcome, without moving any Gradle related files into the root of a project.

Essentially, it's just two lines, wrapping the Android Gradle plugin dependency in the `buildscripts` section:

```
if (project == rootProject) {
    // ... (dependency here)
}
```

With this, the dependency is only evaluated (and possibly downloaded) when the project is opened **stand-alone** (opening the `android` folder in Android Studio).

When a library project is included as a dependency, then `rootProject` will be the main application project. -> the statement will evaluate to false, skipping the potentially outdated Android Gradle plugin download.

I opened a PR with the alternative approach here, which also includes a bit more description: #384 Please check 🙏

PR #384, opened 2019-09-18
Update and fix Gradle setup

friederbluemle:
### Android Gradle plugin

This wraps the Android Gradle plugin dependency in `buildscripts` section in `android/build.gradle` in a conditional:

```
if (project == rootProject) {
    // ... (dependency here)
}
```

A small change with a big impact: 
The Android Gradle plugin is _only_ required when opening the project _stand-alone_, not when it is _included as a dependency_. By doing this, the project opens correctly in Android Studio, and it can also be consumed as a native module dependency from an application project **without** affecting the app project (avoiding unnecessary downloads/conflicts/etc).

Please see related discussion in #381 - This PR replaces and closes #381.

### Gradle wrapper

Similarly the Gradle wrapper (and related scripts) are only required for stand-alone editing/building of the `android` folder. It should _not_ be distributed. Android Studio will automatically generate a wrapper, or alternatively one can be generated on the command line on-the-fly (see below).

### Missing dependency

The `dependencies` section was missing a reference to `androidx.browser:browser:1.0.0`, causing a compile error in `RNAppAuthModule`.

### Missing Android debug keystore in example

Due to an oversight, the `debug.keystore` file was ignored - This is already fixed on react-native's `master` branch: https://github.com/facebook/react-native/blob/master/template/_gitignore#L43
I re-added the keystore to fix the build error in the `AndroidExample` project.

### .npmignore config

In a separate commit, this PR also includes a small update to the npmignore config of the project. By excluding the Gradle wrapper jar and scripts a significant size reduction of the distribution artifact to less than a third of the previous value was achieved:

Before:

```
npm notice === Tarball Details ===
npm notice name:          react-native-app-auth
npm notice version:       5.0.0-rc1
npm notice filename:      react-native-app-auth-5.0.0-rc1.tgz
npm notice package size:  74.6 kB
npm notice unpacked size: 157.0 kB
npm notice total files:   26
```

After:

```
npm notice name:          react-native-app-auth
npm notice version:       5.0.0-rc1
npm notice filename:      react-native-app-auth-5.0.0-rc1.tgz
npm notice package size:  22.8 kB
npm notice unpacked size: 96.1 kB
npm notice total files:   21
```

### What's required for testing (prerequisites)?

Android Studio and/or Gradle installed locally.

### What are the steps to reproduce (after prerequisites)?

Open `android` in Android Studio:

```sh
studio android/
```

Or:

Generate a Gradle wrapper on the fly, and build on command line:

```sh
cd android/
gradle wrapper --gradle-version 5.4.1 --distribution-type all
./gradlew build
```

Successfully verified and tested on:
- Android Device (Xiaomi Mi A2)
- Android Emulator (Pixel 3a)
- iOS Device (iPhone 8)
- iOS Simulator (iPhone X)

<sub><sup>Once approved, please use a *normal* GitHub merge (i.e. **NO rebase/squash** merge) to integrate the commit(s) from the PR head branch. The changes are broken up into meaningful, [atomic commits](https://www.freshconsulting.com/atomic-commits/), and my branch should already be up-to-date with the latest base branch. If it isn't, or if you want me to change anything, please let me know, and I will update the branch as soon as possible. Thank you!</sup></sub>


android/build.gradle
```diff
@@ -3,25 +3,25 @@ def safeExtGet(prop, fallback) {
...
     }
 }
 
 apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 27)
-    buildToolsVersion safeExtGet('buildToolsVersion', '28.0.2')
-
+    buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
```
dulmandakh: it's not needed anymore, so it's better to remove it.

friederbluemle: You're correct, with more recent versions of the Android Gradle plugin this is no longer _required_ - However: At some point (about a year ago), adding this would actually cause a warning in the build log. Also Android Studio templates did not have this line anymore. Then something changed: After another update, the line was **added again** in the Android Studio template (it's even still present in the latest stable version 3.5.0). There is also no more warning, as far as I can tell. Do you happen to know more details by any chance?
Also - Since the version of the Android Gradle plugin is **not** determined by the library project, but by the app project, there is a chance that the app still uses an older version that requires this line (although today this chance is very low, or even non existent). Removing it could break it, while keeping it there can do no harm (other than possibly a warning). Thoughts?

dulmandakh: Play Store requires 64bit support, so developers must use at least RN 0.59. And to have proper AndroidX support, developers use RN 0.60. So there is a zero chance to an app will use older version.

dulmandakh:
Also, I think that we should have gradle wrapper since it maybe be used as a root project.

android/build.gradle
friederbluemle: I understand. But do you know what's the reason this was reverted in the official template (i.e. the line reappeared, and it's no longer a warning)?

kadikraman:
Hi @dulmandakh and @friederbluemle - thanks so much, and I apologise for the delay in getting back you you.

I've looked at both implementations and I'm tempted to go with this one. Are you both in agreement with this or is there more to be done in your opinion @dulmandakh ?

friederbluemle:
Hi @kadikraman - Thanks for the review! :+1:
I rebased the branch, and also updated the Android Gradle plugin to the now latest version (`3.5.1`). As explained, it will only be used when the project is opened stand-alone (it won't affect app projects using this library).

package.json
```diff
@@ -20,13 +20,8 @@
     ]
   },
   "files": [
-    "package.json",
-    "CODE_OF_CONDUCT.md",
-    "CONTRIBUTING.md",
-    "LICENSE",
-    "ios",
     "android",
-    "index.js",
```
kadikraman: Could you add these files back, please? In particular, removing the `index.js` would break deploying new versions (since only the files listed here will get pushed to npm)

package.json
```diff
@@ -69,8 +64,8 @@
     "jest": "^21.2.1",
     "lint-staged": "^6.1.0",
     "prettier": "1.10.2",
-    "react": "^16.1.1",
-    "react-native": "^0.50.3"
+    "react": "16.8.6",
+    "react-native": "0.60.3"
```
kadikraman: Are there any major breaking changes between 60.3, 60.2 and 60.1? If not, I'd prefer this to be `^0.60`.

friederbluemle: There are no major breaking changes that I know of (in the versions you mentioned). However, there no way to guarantee that. The reason I chose `0.60.3` here is to align it with the example project (we can _update_ it independently in a separate PR). The problem with using `^0.60` here is that RN does not strictly follow semver (yet?). `^` fixes the _major_ version only, so it would happily update it to 0.61.* (or 0.62.x once release etc). And those version _might_ contain breaking changes.

package.json
friederbluemle: `index.js` is set as `main` in `package.json`. Just like a bunch of other files, it is _always_ included in the packed version, and there no need to manually specify it. You can just do a quick `npm pack --dry-run` locally to see that it is included even though it's not in the files list.

friederbluemle:
Hi @kadikraman - Just a friendly reminder to please take another quick look at these changes and merge if everything is clear. I rebased the branch again onto the latest `master` branch. Thanks! 🙏

friederbluemle:
Updated to the latest Android Gradle plugin `3.5.2`.

friederbluemle:
Rebased and resolved conflicts.

kadikraman:
Hi @friederbluemle thanks for everything - could you please merge in the latest master (or give me push access to your fork, I've done it locally before realising I couldn't push 🤦‍♀ )

Hoping to get this merged this week.

friederbluemle:
Hi @kadikraman - Thanks for getting back here, and happy new year!

I've rebased the branch and resolved the conflicts. One of the three commits (`Re-add missing Android debug.keystore`) was no longer applicable (since `Example/AndroidExample` was removed), so the commit got removed too.

To fix the following compile error, I had to update the default `compileSdkVersion` from `27` to `28`:

```
appcompat-1.0.2/res/values-v28/values-v28.xml:5:5-8:13: AAPT: error: resource android:attr/dialogCornerRadius not found.
```

The project now builds without issues both on command line and in Android Studio. Both `yarn test` and `yarn lint` pass as well.

kadikraman:

PR #393, opened 2019-10-06
Move to GitHub Actions

sheepsteak:
Migrate to GitHub Actions from Sail CI.

## Description

- Add a GitHub Actions workflow that matches what was in Sail CI file
- Remove Sail CI file as no longer needed

## Steps to verify

You can see the workflow here - https://github.com/sheepsteak/react-native-app-auth/commit/496343aa1fe3e7d37f00dd8bae05cfbe0d7546cf/checks?check_suite_id=253355904

I'm not sure if your org is in the GitHub Actions beta but I think you get instant access now if you go to https://github.com/features/actions

Thanks for trying Sail CI!


sheepsteak:
I've just noticed your using GitHub Actions on https://github.com/FormidableLabs/reason-urql so this should all just work 😅

Issue #291, opened 2019-04-10
Revocation requires a basic auth header in identity server 4


iamkoch:
## Issue

Revocation in identity server 4 requires a basic auth header, which this library seems to omit.

An example request taken from http://docs.identityserver.io/en/latest/endpoints/revocation.html is 

```
POST /connect/revocation HTTP/1.1
Host: server.example.com
Content-Type: application/x-www-form-urlencoded
Authorization: Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW

token=45ghiukldjahdnhzdauz&token_type_hint=refresh_token
```

The library would optionally be configured to pass the client id (and optional secret) into the basic auth header.

---

## Environment

- Identity Server 4
- **Platform that you're experiencing the issue on**:
  - [ ] iOS
  - [ ] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [x] Both
- **Are you using Expo, e.g. `ExpoKit`?**
  - [x] No
  - [ ] Yes, I've _not_ ejected
  - [ ] Yes, but I **have** ejected to `ExpoKit`
  - [ ] Yes, but I **have** ejected to vanilla React Native

kadikraman:
You're correct. An Authorization header should be added for [non-public clients](https://tools.ietf.org/html/rfc6749#section-2.3).

iamkoch:
The request looks like https://tools.ietf.org/html/rfc7009#page-5. This works for public clients as well where secret can be left blank in the Authorization header. I have tested this against identity server 4 for a public PKCE-enabled client with no secret.

PR #407, opened 2019-10-25
Fix/revoke authorization

### LINKED ISSUE #406: A provider might needs an authorization header in the revoke endpoint
## Issue

As described in [rfc7009#section-2.1](https://tools.ietf.org/html/rfc7009#section-2.1) the client can also includes its authentication credentials in the request

---

## Environment

* **Your Identity Provider**: `Custom provider`
* **Platform that you're experiencing the issue on**: `both`
* **Are you using Expo?**: `No`

ehapmgs:
Fixes #406 
Fixes https://github.com/FormidableLabs/react-native-app-auth/pull/291

## Description

adding an authorization header for the revoke endpoint where some provider might need it


index.js
```diff
@@ -167,8 +168,8 @@ export const refresh = (
 };
 
 export const revoke = async (
-  { clientId, issuer, serviceConfiguration },
-  { tokenToRevoke, sendClientId = false }
+  { clientId, issuer, serviceConfiguration, clientSecret },
+  { tokenToRevoke, sendClientId = false, includeBasicAuthorization = false }
```
kadikraman: `includeBasicAuthorization` should also be added to `RevokeConfiguration` in `index.d.ts`

index.js
```diff
@@ -199,6 +203,7 @@ export const revoke = async (
     method: 'POST',
     headers: {
       'Content-Type': 'application/x-www-form-urlencoded',
+      Authorization: `${includeBasicAuthorization ? `basic ${basicAuth}` : ''}`,
```
kadikraman: I'd prefer creating headers as a constant above and adding the auth header optionally:
```
const headers = {
  'Content-Type': 'application/x-www-form-urlencoded',
};

if (includeBasicAuthorization) {
  headers.Authorization =`basic ${base64.encode(`${clientId}:${clientSecret}`)}`,
}
```

This way you only have one conditional statement and it's a bit easier to follow

ehapmgs: done

index.js
ehapmgs: please check a22518e9c1044b6522d0ccf8d0ceba4eee4d2884

index.js
kadikraman: Thank you! I've just been thinking about this PR, and would really like to avoid adding a dependency if possible (I'm planning on getting rid of `invariant` as well).

What do you think about instead of `includeBasicAuthorization`, we could pass in a `customHeaders` param (we already do something similar for the `authorize` call). This way, the logic for the base 64 encoding would live in the app code, rather than the library 🤔

ehapmgs: I had a similar idea at first but I thought we could make simpler for the app code since we already have clientId & clientSecret in the config (I guess the native library is doing the same), but the down side of it we would add an extra dependency.

let me know what you think I will be happy to create a new pull request for your approach.

kadikraman: Alright, let's just leave it as 👍

kadikraman:

PR #412, opened 2019-10-29
Typescript: Add null as posible token value for RefreshResult

henninghall:
At least one of the providers (Google) returns no new refresh token upon refresh. Instead, the previous one can be used again.


kadikraman:

PR #415, opened 2019-11-08
Add call to beginBackgroundTaskWithExpirationHandler before presenting authorization request

### LINKED ISSUE #414: Two-factor authentication that requires user to switch to different app fails on iOS13.
## Issue

If the login flow includes two factor authentication (2FA) many times a user is required to switch to a different app to confirm the second form of authentication such as finding a text message with a code that was sent or confirming a push notification in an app like Symantec VIP Access.  This flow often requires the form where the user initiated the second factor of authentication to initiate a long-running network request waiting for the user to confirm the authentication request. 

In iOS13 the network request in the AsWebAuthenticationSession control is killed when the user puts the app in the background.  When the user resumes the app the webview displayed in the control has an error that reads "Safari cannot open the page because the network connection was lost."  In iOS12 this is not an issue.

The link below includes videos illustrating the problem and confirming it was not an issue in iOS12.  One video file is titled iOS13 and another is title iOS12.  Both videos show an active long running network connection initiated (simulating a request waiting for a user to confirm authentication) and then the app is put into the background and resumed a few seconds later.  In iOS13 the error is displayed on resume, in iOS12 the network request continues to run.

https://drive.google.com/drive/folders/1vvAEUUyGTOSx9HI5GPVNYIAX5e4KZBrm

Here's another set of videos showing an actual 2-factor request initiating from Ping where the user switches over to Symantec VIP app to confirm push notification and when resuming the app the error is displayed.  There are also videos that show if the user confirms the push using force touch, thus keeping the original app in the foreground, the network request completes successfully.  It also has a video that shows if the user uses text message to confirm identity (so no network request is initiated before the user switches apps), the flow works fine.

https://drive.google.com/drive/folders/1yCZoizMuVMZ6CFov87ojEZrkyyLnBfpZ

## Environment

* **Your Identity Provider**: Ping with Symantec for 2FA
* **Platform that you're experiencing the issue on**: iOS13
* **Are you using Expo?**: No

georgeloring:
Fixes #414

## Description

Adding a call to [beginBackgroundTaskWithExpirationHandler](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtaskwithexpiratio) before presenting AsWebAuthenticationSession control so that if a network request is in progress in iOS13 and the app is put into the background, the network request is not killed right away.  Background task will still be expired if takes too long or directly expired when the authentication session completes.

## Steps to verify

See video links from bug report.  With a long-running network request in AsWebAuthenticationSession put the app in the background and then resume.  With this fix, the network request is still intact.

### DIFF
--- a/ios/RNAppAuth.m
+++ b/ios/RNAppAuth.m
@@ -24,6 +24,8 @@
     return dispatch_get_main_queue();
 }
 
+UIBackgroundTaskIdentifier taskId;
+
 /*! @brief Number of random bytes generated for the @ state.
  */
 static NSUInteger const kStateSizeBytes = 32;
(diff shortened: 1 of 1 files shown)


kadikraman:

TheMatrixMaster:
### Description
This issue still persists on iOS14. We are using the latest react-native-app-auth version (v6.2.0). Following a standard Azure Active Directory v2 authentication request, the user leaves our app to validate two-factor authentication in Microsoft Authenticator app. This will resolve the request and return the appropriate access tokens which we then want to send to our server. However, the network request that gets created when the app is in the background gets killed immediately instead of staying intact until the user re-accesses the app.

### Verification
This video link shows the described behaviour above.
https://drive.google.com/file/d/1LMs6t76iSaTyu4vdR5Gk5ZqaKsjwJCEe/view?usp=sharing

Issue #426, opened 2019-12-19
onActivityResult crashing when intent is null because promise is null


Shervanator:
## Issue

This is related to the fix here: https://github.com/FormidableLabs/react-native-app-auth/pull/352 and the initial issue here: https://github.com/FormidableLabs/react-native-app-auth/issues/130

When `onActivityResult` is called (e.g. when returning to app from another app) the intent can be null. This triggers the following flow:
https://github.com/FormidableLabs/react-native-app-auth/blob/cffea61208a7b37a674d4a4be25fc9c70c95e4b2/android/src/main/java/com/rnappauth/RNAppAuthModule.java#L302-L305

Which try's to do the following:
```
promise.reject("Failed to authenticate", "Data intent is null" );
```

However if none of the other exported functions in this module are called then the `promise` instance variable is null. Because the `promise` is only set when the other functions are called. e.g. https://github.com/FormidableLabs/react-native-app-auth/blob/cffea61208a7b37a674d4a4be25fc9c70c95e4b2/android/src/main/java/com/rnappauth/RNAppAuthModule.java#L158

I think we should check if the promise is set before calling it in the `onActivityResult` function? 

cc @AlgirdasVZ @kadikraman 

I'm going to play around with a fix since this is effecting us in a production app at the moment! 

---

## Environment

* **Your Identity Provider**: `*`
* **Platform that you're experiencing the issue on**: `Android`
* **Are you using Expo?**: `no`
* **Package version?**: `5.0.0-rc3`

Shervanator:
I've put together a PR that fixes the issue here: https://github.com/FormidableLabs/react-native-app-auth/pull/427 would love to get this merged so we don't have to maintain a forked package! :D

AlgirdasVZ:
Hi, can you post here the full crash report of this issue?

Shervanator:
hello! this is the crash:
```
12-19 08:51:45.417  4671  4671 E AndroidRuntime: FATAL EXCEPTION: main
12-19 08:51:45.417  4671  4671 E AndroidRuntime: Process: com.envato.development.Milkshake, PID: 4671
[...]

PR #427, opened 2019-12-19
check promise before calling it in the android onActivityResult function

Shervanator:
Fixes https://github.com/FormidableLabs/react-native-app-auth/issues/426

## Description

`onActivityResult` is called by all intents when returning to the app from some other location. In the case that we are returning to the app (but not for app auth/or when the intent is null) we trigger a new flow that was added here: https://github.com/FormidableLabs/react-native-app-auth/pull/352
This new flow expects the `promise` instance variable to be set. This only happens in some of the other functions in the native module, e.g. the `authorize` function.

This PR ensures that we check if the promise is defined before we call it. Otherwise the app will crash (which is what I was experiencing prior to this PR)

cc @AlgirdasVZ @kadikraman


kadikraman:
Thanks for the PR! Happy to merge if you could resolve the conflicts (the error messaging has been changed so we could have consistent messaging across platforms)

kadikraman:
Nevermind - looks like I can resolve the conflicts through GitHub UI ✔️

kadikraman:

PR #470, opened 2020-02-26
Fix android multiple providers

### LINKED ISSUE #332: Configuration is cached between authorize calls
## Issue

When calling authorize method with different configuration, on the second time the configuration is taken from the first call - it is cached after first call.

The problem exist in version v4.4.0, but not in v4.3.1. I believe that the issue was introduced in this commit:
https://github.com/FormidableLabs/react-native-app-auth/commit/44a47c524dd20ec47595ece6730f650d70ac6078

The mServiceConfiguration is taken instead of serviceConfiguration, which is ok for the first time or for using prefetching, and the field is not cleared when closing the chrome tab.

Steps to reproduce:
1. call authorize with issuer X
2. do not log in, just close the tab
3. call authorize with issuer Y

**Result:** tab is opened with issuer X
**Expected result:** tab is opened with issuer Y

---

## Environment

* **Your Identity Provider**: `my own`
* **Are you using Expo?**: `no`

aeirola:
Fixes #332

## Description

Fixes issue on Android where the wrong authorization endpoint would be opened when changing authroziation provider. This is done by adding support for caching service configurations for different issuers.

## Steps to verify

Run application in android, and open the different authentication providers. Check that each opens the correct authentication providers authorization page.

### DIFF
--- a/Example/App.js
+++ b/Example/App.js
@@ -1,6 +1,6 @@
 import React, { useState, useCallback, useMemo } from 'react';
 import { UIManager, LayoutAnimation, Alert } from 'react-native';
-import { authorize, refresh, revoke } from 'react-native-app-auth';
+import { authorize, refresh, revoke, prefetchConfiguration } from 'react-native-app-auth';
 import { Page, Button, ButtonContainer, Form, FormLabel, FormValue, Heading } from './components';
 
 UIManager.setLayoutAnimationEnabledExperimental &&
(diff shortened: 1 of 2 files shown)


Example/App.js
```diff
@@ -1,84 +1,114 @@
...
+    additionalParameters: {},
+    scopes: ['openid', 'profile', 'email', 'offline_access'],
+
+    // serviceConfiguration: {
+    //   authorizationEndpoint: 'https://demo.identityserver.io/connect/authorize',
+    //   tokenEndpoint: 'https://demo.identityserver.io/connect/token',
+    //   revocationEndpoint: 'https://demo.identityserver.io/connect/revoke',
+    // },
+  },
+  auth0: {
```
aeirola: These changes to the example are optional, but make testing the changes easier.

kadikraman:
Released in [v5.1.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v5.1.0) 🎉

PR #475, opened 2020-02-28
Fix android crashing on null issuer

### LINKED ISSUE #473: [Android] Error `Attempt to invoke virtual method 'int java.lang.Object.hashCode()' on a null object reference` when call authorize
## Issue

---
```js
const f = async() => {
  try {
    const ret = await authorize(config)
    console.log(ret)
   } catch (error) {
     console.log('Failed to log in', error.message)
   }
 }
 f()
```

Error `Attempt to invoke virtual method 'int java.lang.Object.hashCode()' on a null object reference` when call `authorize(config)`

## Environment

* **Your Identity Provider**:  It works with iOS, but not Android
* **Platform that you're experiencing the issue on**: `only Android `
* **Are you using Expo?** No

aeirola:
Fixes #473

## Description

#470 caused a regression where a missing `issuer` field in the configuration would cause calls to fail. This PR updates the code to handle the cases where issuer is null.

## Steps to verify

Add `serviceConfiguration` and remove `issuer` configurations and verify that authorization still works under android.

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -97,15 +97,15 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
         final CountDownLatch fetchConfigurationLatch = new CountDownLatch(1);
 
         if(!isPrefetched) {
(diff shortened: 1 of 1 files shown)


kadikraman:
Good catch, thank! Release this as a patch in [v5.1.1](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v5.1.1)

PR #482, opened 2020-03-12
Expose OAuth error codes

### LINKED ISSUE #474: Identify revoked tokens during token
Our app would like to offer the user the possibility to reauthenticate in cases where the refresh token has been revoked on the server side. Currently it is difficult to determine the reason for a failure during token refresh. For example both a network failure and a revoked refresh token throw an error with the code `token_refresh_failed`, making it hard to distinguish between the two scenarios.

Now, it seems as if the error message when the refresh token has been revoked starts with `invalid_grant: `on iOS, so it would be possible to try and parse that content. But that is not the case on Android, and even if that would be the case it would be undocumented territory and easily subject to change.

So would be useful to have more detailed information in the errors, especially during token refresh where various failures can mean vastly different things. This could maybe be accomplished by extending the existing [error codes](https://github.com/FormidableLabs/react-native-app-auth#error-messages), or adding a separate field for the error cause, which could specify the different known scenarios.

aeirola:
Fixes #474

## Description

Exposes OAuth error codes to the JavaScript client.

## Steps to verify

Add deliberate errors in the OAuth calls, log the error responses and check that the appropriate code is present in the error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,11 @@ try {
 
 ## Error messages
 
+Values are in the `code` field of the rejected Error object.
+
+- OAuth Authorization [error codes](https://tools.ietf.org/html/rfc6749#section-4.1.2.1)
+- OAuth Access Token [error codes](https://tools.ietf.org/html/rfc6749#section-5.2)
+- OpendID Connect Registration [error codes](https://openid.net/specs/openid-connect-registration-1_0.html#RegistrationError)
 - `service_configuration_fetch_error` - could not fetch the service configuration
 - `authentication_failed` - user authentication failed
 - `token_refresh_failed` - could not exchange the refresh token for a new JWT
--- a/ios/RNAppAuth.m
+++ b/ios/RNAppAuth.m
@@ -243,7 +243,8 @@ RCT_REMAP_METHOD(refresh,
                                                  if (response) {
                                                      resolve([self formatRegistrationResponse:response]);
                                                  } else {
-                                                     reject(@"registration_failed", [error localizedDescription], error);
+                                                     reject([self getErrorCode: error defaultCode:@"registration_failed"],
+                                                            [error localizedDescription], error);
                                                  }
                                             }];
 }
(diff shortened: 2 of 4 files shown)


ios/RNAppAuth.m
```diff
@@ -407,4 +410,51 @@ - (NSDictionary*)formatRegistrationResponse: (OIDRegistrationResponse*) response
              };
 }
 
+- (NSString*)getErrorCode: (NSError*) error defaultCode: (NSString *) defaultCode {
```
aeirola: Not an Objective-C programmer, so just copied the example from Apple documentation https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-BAJCIGIA. Sorry if this is against some conventions used by the library.

README.md
```diff
@@ -395,6 +395,11 @@ try {
 
 ## Error messages
 
+Values are in the `code` field of the rejected Error object.
+
+- OAuth Authorization [error codes](https://tools.ietf.org/html/rfc6749#section-4.1.2.1)
```
aeirola: **Note!** Adding more possible error codes here might be considered a breaking change according some conventions. For example someone expecting some errors to always include `code: 'token_refresh_failed'` might be disappointed when discovering that a new version of the library replaced the value with `code: 'invalid_grant'` in some situations.

If we don't want to use the same `code` field for the new values, we could move them to a separate field such as `oauthCode`. This would require some extra work though to move the value through the promise rejection call.

kadikraman:
Thank you so much for this PR! This definite improvement over the existing solution, but like you said, it's a breaking change, since, it's likely that there are users relying on the error message being one of the pre-defined strings.

I'm keen to get this merged because surfacing the error codes will be very helpful for debugging, but it'll have to go into the v6.0.0 release.

README.md
kadikraman: Yeah it's definitely a breaking change currently. As tempting as it is to move the auth code to an `oauthCode` field, I think the "right" thing to do is to surface the actual error codes and do a major release for it.

PR #498, opened 2020-04-13
Update platform iOS to 10.0 in podspec

suraneti:
# Description

According to React core and React podspec right now is `10.0` should it better to set the same version, it will resolve warning about a target version mismatch.

[React-Core.podspec](https://github.com/facebook/react-native/blob/master/React-Core.podspec)
[React.podspec](https://github.com/facebook/react-native/blob/master/React.podspec)

## Steps to verify

1. `yarn add react-native-app-auth`
2. `cd ios && pod install`

### DIFF
--- a/react-native-app-auth.podspec
+++ b/react-native-app-auth.podspec
@@ -9,7 +9,7 @@ Pod::Spec.new do |s|
   s.license      = package['license']
   s.authors      = package['author']
   s.homepage     = package['homepage']
-  s.platform     = :ios, '9.0'
+  s.platform     = :ios, '10.0'
   s.source       = { :git => 'https://github.com/FormidableLabs/react-native-app-auth.git' }
   s.source_files  = 'ios/**/*.{h,m}'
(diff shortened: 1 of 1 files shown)


jpdriver:
thanks for opening this issue @suraneti 

it looks like right now the [change to the React podspecs](https://github.com/facebook/react-native/commit/674b591809cd1275b5f1c4d203c2f0ec52303396) is awaiting release, and will only warn if you're using either an RC version of `react-native@0.63.x` or building straight from `master`

i don't think we should jump ahead of the iOS version the stable React Native release currently supports, as doing so would force anyone using this library with RN 0.62.x into deprecating iOS 9 support from their app early 😱 

let's wait until 0.63 has been released officially, and then we can absolutely merge this change

jpdriver:
updated to note that RN 0.63 has now shipped so we can merge this when we're ready to start prepping the next release.

@kadikraman can we include this in the next major?

PR #508, opened 2020-05-10
Return scopes as string

r0b0t3d:
## Description
In Strava, they return scopes as `read,activity:read`. Current code can not parse it because it used `space` as a separator.
```
String[] scopesArray = scope.split(" ");
```
```
@"scopes": authResponse.scope ? [authResponse.scope componentsSeparatedByString:@" "] : [NSArray new],
```

## Solution
I suggest to return scopes as string instead. So it will works in every cases.

## Breaking changes
Scopes returned as string instead of array

Before
```
{
  "accessToken": "",
  "accessTokenExpirationDate": "",
  "authorizeAdditionalParameters": {},
  "idToken": null,
  "refreshToken": "",
  "scopes": [],
  "tokenAdditionalParameters": {},
  "tokenType": "Bearer"
}
```
After
```
{
  "accessToken": "",
  "accessTokenExpirationDate": "",
  "authorizeAdditionalParameters": {},
  "idToken": null,
  "refreshToken": "",
  "scopes": "read,activity:read",           <-- This line
  "tokenAdditionalParameters": {},
  "tokenType": "Bearer"
}
```

---

will fix https://github.com/FormidableLabs/react-native-app-auth/issues/522

### DIFF
--- a/android/src/main/java/com/rnappauth/utils/TokenResponseFactory.java
+++ b/android/src/main/java/com/rnappauth/utils/TokenResponseFactory.java
@@ -57,7 +57,7 @@ public final class TokenResponseFactory {
         map.putString("idToken", response.idToken);
         map.putString("refreshToken", response.refreshToken);
         map.putString("tokenType", response.tokenType);
-        map.putArray("scopes", createScopeArray(authResponse.scope));
+        map.putString("scopes", authResponse.scope);
 
         if (response.accessTokenExpirationTime != null) {
             map.putString("accessTokenExpirationDate", DateUtil.formatTimestamp(response.accessTokenExpirationTime));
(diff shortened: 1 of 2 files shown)


jpdriver:
hey @r0b0t3d -- thanks for opening this issue. that's an annoying choice by Strava 🙄 do you know any other providers that also use comma-separated scopes? from a couple of clicks around, it looks like the majority use space-separated ones..

we do need to support `scopes` for Strava; but i'm a little hesitant to introduce a breaking change in this library just to support a small handful of providers.

how about instead we extend the existing `createScopeArray` method (and perhaps introduce a similar transform on the iOS side) so that we can support both delimiters depending on which one is present?

personally, i think we should keep the return type as an Array so there won't be a breaking change required in user-land to add this support in..

what do you think? 🙂

r0b0t3d:
@jpdriver just tried `Strava` for now.
If most of the providers use space-separated, I think the best way to not introduce breaking changes here.
I will try your suggestion to support 2 kind of separators (space and comma).

PR #518, opened 2020-06-11
Fix crash on android when warming up custom tabs

### LINKED ISSUE #517: Android custom tabs warmup crashes on some Samsung devices
## Issue

When using the `warmAndPrefetchChrome` configuration on Android Samsung devices it is possible that the app crashes with the error `java.lang.NullPointerException: Attempt to invoke virtual method 'boolean CustomTabsSession.newSession (android.net.Uri, android.os.Bundle, java.util.List)' on a null object reference`.

This seems to be due to a missing `null` check in [`onCustomTabsServiceConnected`](https://github.com/FormidableLabs/react-native-app-auth/blob/v5.1.2/android/src/main/java/com/rnappauth/RNAppAuthModule.java#L765). The `CustomTabsSession.newSession` [documentation](https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsClient#newSession(androidx.browser.customtabs.CustomTabsCallback,%2520int)) states that the method may return null on errors. These errors shouldn't crash the app though.

A similar issue was present, and fixed in https://github.com/openid/AppAuth-Android/issues/315

---

## Environment

* **Your Identity Provider**: `IdentityServer 4`
* **Are you using Expo?** `no`

aeirola:
Fixes #517

## Description

Handles possibly `null` values returned by [`CustomTabsSession.newSession`](https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsClient#newSession(androidx.browser.customtabs.CustomTabsCallback,%2520int)).

## Steps to verify

1. Have an Android Samsung device
2. Prefetch configuration with `warmAndPrefetchChrome` enabled
3. Should not crash the app

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -12,6 +12,7 @@ import androidx.annotation.Nullable;
 import androidx.browser.customtabs.CustomTabsCallback;
 import androidx.browser.customtabs.CustomTabsClient;
 import androidx.browser.customtabs.CustomTabsServiceConnection;
+import androidx.browser.customtabs.CustomTabsSession;
 
 import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.ReactApplicationContext;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -762,7 +763,11 @@ private void warmChromeCustomTab(Context context, final String issuer) {
             @Override
             public void onCustomTabsServiceConnected(ComponentName name, CustomTabsClient client) {
                 client.warmup(0);
```
mikehardy, line 765: Did you know that .warmup can throw? We experienced crashes here before: https://github.com/ankidroid/Anki-Android/commit/6503e5dda718397a41de9c90e14a0a7ded0e0e37

https://github.com/ankidroid/Anki-Android/pull/6146/files#diff-b4eaf9bae4ba5cd94138b46ef441e244R154

You have to catch IllegalStateException + SecurityException, or just Exception really

The crashes are rare but they do come in

PR #521, opened 2020-06-24
[Android] Authenticate using third party app

r0b0t3d:
## Issue
Always use webview for authenticating

## Expected
If `Strava` app installed. Open the app instead of webview

## Description
Add option to pass the third party package name. First, check if this app is installed, open app. Otherwise, use webview as before

## Steps to verify

1. Install Strava app
2. Add option `use_app` to the `additionalParameters `
```
const config: AuthConfiguration = {
        serviceConfiguration: {
          authorizationEndpoint: ''
          tokenEndpoint: '',
          revocationEndpoint: 'https://www.strava.com/oauth/deauthorize',
        },
        redirectUrl,
        scopes: ['activity:read,activity:read_all'],
        additionalParameters: {
          appOAuthUrlScheme: 'strava://oauth/mobile/authorize',
          use_app: 'com.strava',
        },
      };
```
3. Start authenticating with Strava inside our app.
4. Strava app launched for authenticating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ with optional overrides.
 - **additionalParameters** - (`object`) additional parameters that will be passed in the authorization request.
   Must be string values! E.g. setting `additionalParameters: { hello: 'world', foo: 'bar' }` would add
   `hello=world&foo=bar` to the authorization request.
+  - `use_app`: [Android only] specific the application package name that you'd like to use for authenticating. Otherwise, use webview as default.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -120,6 +120,7 @@ with optional overrides.
 - **additionalParameters** - (`object`) additional parameters that will be passed in the authorization request.
   Must be string values! E.g. setting `additionalParameters: { hello: 'world', foo: 'bar' }` would add
   `hello=world&foo=bar` to the authorization request.
+  - `use_app`: [Android only] specific the application package name that you'd like to use for authenticating. Otherwise, use webview as default.
```
kadikraman, line 123: We can't use `additionalParameter` passing adding properties to the auth function. `additionalParameters` is is a key-value map of values to be passed to the auth request. This would have to be a top lever parameter (more like `dangerouslyAllowInsecureHttpRequests`).

Issue #401, opened 2019-10-18
App freezes when calling prefetchConfiguration with no internet connection


mtroskot:
## Issue

When calling prefetchConfiguration with device in airplane mode or with no internet connection the app freezes.

This is the error I get:  E/AppAuth: Network error when retrieving discovery document.

config.js file
```js
import { prefetchConfiguration } from 'react-native-app-auth';

const config = {
  warmAndPrefetchChrome: true,
  issuer: 'https://login.microsoftonline.com/1caa43b8-bf09-48b6-9b3c-bd5a56fec019',
  clientId: '5e0b7c9c-aaaa-472a-8ec6-6362c6d56fbe',
  redirectUrl: 'msauth.com.rccl.oauth://auth/',
  additionalParameters: {},
  scopes: ['openid', 'profile', 'email', 'offline_access'],
};

prefetchConfiguration(config);

export default config;
```

Also tried using try/catch, still the app freezes.
```js
try{
prefetchConfiguration(config);
}catch (error) {
  console.log(error);
} 
```

## Environment
RN: 0.59.4
react-native-app-auth: ^4.4.0

* **Your Identity Provider**: Azure
* **Platform that you're experiencing the issue on**: both
* **Are you using Expo?** No

e3bmo3ty:
any fix ?

kadikraman:
I think in general we should check on the React Native side whether the app is connected to the internet before triggering this request, since none of the methods would work with no internet... but it's probably sensible to have a fallback, so we could add a timeout to the request.

Vikasatneaxus:
any one have fixed this error
[Error: Network error]

tomjoj:
use in config
dangerouslyAllowInsecureHttpRequests: true,

PR #545, opened 2020-08-05
Bugfix/custom tabs timeout

kadikraman:
Fixes https://github.com/FormidableLabs/react-native-app-auth/issues/401

## Description

On Android, `warmAndPrefetchChrome` makes the app not load if the device is not connected to the internet. This adds a default timeout to the request, and a timeout prop to adjust the timeout in seconds.

## Steps to verify

1. Load the example app (in this repository) using the version of the library from this PR
2. Open the app in an Android emulator
3. Disconnect the emulator from internet
4. Refresh the app and ensure it loads
5. Check that the timeout prop passed into `prefetchConfiguration` changed how long it takes for the page to load

### DIFF
--- a/Example/App.js
+++ b/Example/App.js
@@ -1,5 +1,5 @@
-import React, { useState, useCallback, useMemo } from 'react';
-import { UIManager, Alert } from 'react-native';
+import React, { useState, useCallback, useMemo, useEffect } from 'react';
+import { Alert } from 'react-native';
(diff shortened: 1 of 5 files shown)


kadikraman:
Closing this since after a discussion with @jpdriver we've decided that this wouldn't solve the underlying problem.

Long term, we could add a check in the JS side of things to verify that the device is connected to the internet before calling any of the methods. However, the immediate fix would be to simply instruct users to do this check on their side before calling any of the methods. I'll update the example app to do this for now.

PR #562, opened 2020-09-22
Fix React dependency to enable XCode 12 build

harunsmrkovic:
Fixes https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

## Description

Check linked Issue, it describes the faulty dependency.

### DIFF
--- a/react-native-app-auth.podspec
+++ b/react-native-app-auth.podspec
@@ -13,6 +13,6 @@ Pod::Spec.new do |s|
   s.source       = { :git => 'https://github.com/FormidableLabs/react-native-app-auth.git', :tag => "v#{s.version}" }
   s.source_files  = 'ios/**/*.{h,m}'
   s.requires_arc = true
-  s.dependency 'React'
+  s.dependency 'React-Core'
   s.dependency 'AppAuth', '1.4.0'
 end


jpdriver:
hey @harunsmrkovic -- thanks for picking this up!

both the Example app and a completely fresh RN 62.3 app build fine for me either with or without this change (Xcode 12A7209) so to that end I think we should merge this in but I don't think there's a huge need for it to be released immediately just yet?

let me know if this is actually causing a build issue for you, or if you're just being a good Open Source citizen 😄 

cheers

harunsmrkovic:
Hey @jpdriver, thanks for the quick merge! yeah, actually lack of this PR is causing a build issue for me, but I'm getting it for a ton of other dependencies as well, so I needed to resort to `patch-package` anyway.

In that context, I don't think that release is urgent, as anyone experiencing this issue will have to resolve it manually anyway 🙃

Guuz:
I'm running into this issue too. And `react-native-app-auth` seems to be (one of) the last package for my project to have not fixed/released this yet. So a release would be welcome!  @jpdriver @harunsmrkovic

jpdriver:
@harunsmrkovic @Guuz this has finally gone out in v6.0.2 🥳

PR #587, opened 2020-12-01
Add custom state as an additional parameter in Android.

### LINKED ISSUE #504: Implement support for "state" parameter
For the iOS version there is behind the scenes state a random state generator, as well as validation at the end of the requests to authorise users.

For the Android version "authRequestBuilder.setState("myStateVar")" is not implemented at all in RNAppAuthModule.

What are the reasons not to support this equally and having the ability to pass by a custom "state"?

klandell:
Fixes #504, adds the ability to add a custom `state` as an additional parameter in Android.

## Description

Implemented `authRequestBuilder.setState` alongside the other special additional parameters.  Per the [AuthorizationRequest.Builder documentation](https://openid.github.io/AppAuth-Android/docs/latest/net/openid/appauth/AuthorizationRequest.Builder.html#setState-java.lang.String-), not explicitly setting the state via this method already adds a random state value.  So this change simply gives the developer a way to control the value.

## Steps to verify

Add a `state` value to the additional parameters object in the configuration in Android.  An error is no longer thrown.

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -547,6 +547,10 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
                 authRequestBuilder.setPrompt(additionalParametersMap.get("prompt"));
                 additionalParametersMap.remove("prompt");
             }
+            if (additionalParametersMap.containsKey("state")) {
+                authRequestBuilder.setState(additionalParametersMap.get("state"));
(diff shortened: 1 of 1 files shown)


klandell:
Hello!  Has there been any consideration in merging this pull request?

kadikraman (approved):
Nice one, thanks so much for adding this!

kadikraman:
This is now released in [v6.0.2](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v6.0.2) 🚀

Issue #380, opened 2019-09-11
Authorize never returns after async call


Highspeed7:
## Issue
I have been trying to implement a solution where the "Reset Password" option is handled when a user is attempting to authorize with my app. I came up with the below, but when this executes the authorize call never returns. It doesn't return with a success or error. I have also noticed this will randomly happen during development when I'm reloading the app several times. It will enter my auth function and never return, thus hanging until I clear cache, clear data and restart the app.
```
export const auth = () => {
    return async(dispatch) => {
        var resetEndpoint = "https://login.microsoftonline.com/giftwizit.onmicrosoft.com/oauth2/v2.0/authorize?p=B2C_1_secretreset1"
        dispatch(authStart())
        try {
            console.log("in auth function");
            const authState = await authorize(authCfg.config);
            // Store token info in asyncStorage
            dispatch(authStoreToken(authState));
        }catch(error) {
            // dispatch(authFail(error));
            if(error.message.indexOf("AADB2C90118") !== -1){
               await authorize({
                   ...authCfg.config,
                   serviceConfiguration:{
                       ...authCfg.config.serviceConfiguration,
                       authorizationEndpoint: resetEndpoint
                   }
               }).then((result) => {
                   console.log(result);
               }).catch((e) => {
                   console.log(e);
               });
               console.log("Password reset complete");
            }else {
                console.log(error);
            }
        }
    }
}
```

## Environment

* **Your Identity Provider**: `e.g. IdentityServer 4 / Okta / Azure`
Azure

* **Platform that you're experiencing the issue on**: `iOS / Android / both`
Android

* **Are you using Expo?**
No

Highspeed7:
I want to add that in most, if not all occurrences that the authorize never returning seems to happen only when the debugger is connected.

sboyd17:
I am having a similar issue. It returns to the app, but code below the call is never executed.

shamprasadrh:
I am also having the same issue

anmour:
I`m also having the same issue!  please help :(

Libanseerefine:
I am also having the same issue with both Refresh and Authorize.
Is there anyone who can confirm this has happened outside of development?

only seen it when I reload to many times and I can also recreate it while reloading while calling the functions.

isnifer:
Same here

isnifer:
@kadikraman please, take a look

ptc-wguo:
> Same here
> 
[...]

PR #593, opened 2020-12-22
Added clarification to the azure example config.

mattwoberts:
## Description

I've added a clarification note to the example azure config. There is a very annoying issue on IOS where if you don't add a trailing slash to the azure `redirtUrl` it won't work - the authorize call will never return.

Hat-top to @isnifer for pointing this out here: https://github.com/FormidableLabs/react-native-app-auth/issues/380#issuecomment-581517451


kadikraman (approved):
Thanks so much for taking the time to add this to the docs! 🙌

Issue #385, opened 2019-09-18
Calls to .authorize or .refresh cause application crash when (and not only) no browser is detected


skippone:
## Issue
Currently native calls to `.authorize()` and `.refresh()` methods on **Android** platform can cause RN application to crash with native exception. Such native exceptions are by definition not catchable in RN app, unless one uses global [native exception handler](https://github.com/master-atul/react-native-exception-handler) and even than the user experience (not mentioning developer's) is rather dreadfull.

For example when underlaying AppAuth library is unable to find suitable browser it throws `ActivityNotFoundException` during `performAuthorizationRequest()` call. Such is the case with certain Android devices with [Opera browser installed ](https://github.com/openid/AppAuth-Android/issues/452) or when browser is [unavailable/disabled](https://github.com/openid/AppAuth-Android/issues/459).

Example of a stack trace for such an application crash when invoking `.authorize()` method in RN application with Opera installed (see issues above):
```
android.content.ActivityNotFoundException:
  at net.openid.appauth.AuthorizationService.prepareAuthorizationRequestIntent (AuthorizationService.java:372)
  at net.openid.appauth.AuthorizationService.getAuthorizationRequestIntent (AuthorizationService.java:267)
  at net.openid.appauth.AuthorizationService.getAuthorizationRequestIntent (AuthorizationService.java:296)
  at com.rnappauth.RNAppAuthModule.authorizeWithConfiguration (RNAppAuthModule.java:326)
  at com.rnappauth.RNAppAuthModule.access$000 (RNAppAuthModule.java:43)
  at com.rnappauth.RNAppAuthModule$1.onFetchConfigurationCompleted (RNAppAuthModule.java:119)
  at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute (AuthorizationServiceConfiguration.java:364)
  at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute (AuthorizationServiceConfiguration.java:305)
  at android.os.AsyncTask.finish (AsyncTask.java:695)
  at android.os.AsyncTask.-wrap1 (Unknown Source)
  at android.os.AsyncTask$InternalHandler.handleMessage (AsyncTask.java:712)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:164)
[...]

PR #599, opened 2021-01-21
Avoid crash on Android when no browsers are found

carrybit:
Additional fix for #385

This is meant as a bugfix on v5 as well as v6.

## Description

A try-catch has been added around the call to _authorizeWithConfiguration_ inside _onFetchConfigurationCompleted_.

Previous fix (#385) handled the situation where the service configuration was passed in manually, while this new fix also handles the case where the configuration is retrieved from the OpenID well-known discovery endpoint.

## Steps to verify

Remove all browsers on Android device. 

Call _authorize_ without setting the _serviceConfiguration_, so that the service configuration will be retrieved from the well-known endpoint.

A browser_not_found error is returned in the promise, instead of throwing an exception which crashes app.

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -270,15 +270,21 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
 
                             setServiceConfiguration(issuer, fetchedConfiguration);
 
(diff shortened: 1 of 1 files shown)


jpdriver (approved):

kadikraman:
Published in [v6.1.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v6.1.0) 🚀

PR #604, opened 2021-02-10
Return codeVerifier value when both skipCodeExchange=true and usePKCE=true

davemurphysf:
## Description
The purpose of this PR is to expose/return the `codeVerifier` value that is used in the request when authenticating against a service that uses/requires PKCE and an authorization code is going to be being passed to a server to exchange for tokens.  Both Google and Microsoft recommend using PKCE, and if you do, they require passing the `code_verifier` in the code exchange leg.  For both Android and iOS, I implemented minimal changes to either create and pass the codeVerifier down to the existing library (Android) or just return/expose a value that was already being used (iOS).


kadikraman (approved):
Thanks a lot adding this!

kadikraman:
Published in [v6.1.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v6.1.0) 🚀

PR #607, opened 2021-03-01
Feature/custom http headers on ios

carrybit:
# Custom HTTP headers on iOS

## Description

Adds functionality for setting HTTP headers to be sent for authorize, register, and refresh requests on iOS.

The same HTTP headers are sent for all of the requests, which means it does not have the full custom experience as we have on Android, where the HTTP headers can be customized individually for each of the requests. Nevertheless, this is a step forward compared to now, as this will enable sending custom HTTP headers also on iOS.

Feature was originally implemented as a private patch to v5, to meet a requirement in a project I am currently working on, and this pull request has been created from that patch.

### DIFF
--- a/index.spec.js
+++ b/index.spec.js
@@ -41,7 +41,7 @@ describe('AppAuth', () => {
     useNonce: true,
     usePKCE: true,
     customHeaders: null,
-    additionalHeaders: { header: "value"},
+    additionalHeaders: { header: 'value'},
     skipCodeExchange: false,
   };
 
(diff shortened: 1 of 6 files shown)


index.spec.js
```diff
@@ -407,11 +447,121 @@ describe('AppAuth', () => {
...
+              ...config,
+              additionalHeaders: {
+                notString: {},
+              },
+            });
+          }).toThrow();
+        });
+      });
+
+      describe('useNonce parameter', () => {
```
carrybit, line 492: The tests for _nonce_ and _usePKCE_ has been grouped into the _authorize_ tests, as they were previously put amongst the _refresh_ tests.

kadikraman (approved):
Fantastic, thanks so much for taking the time to add this! 👏

masiddee:
Thanks again @carrybit! This is now released in v6.2.0 🚀

carrybit:
Sweet!

PR #628, opened 2021-04-27
Update README.md

siddsarkar:
Update readme with unsplash inclusion.


kadikraman:
Hi @siddsarkar  - we don't have a config example for Unsplash, so that "config example" link won't lead to anywhere. Would you like to add a config example as well? :)

siddsarkar:
> Hi @siddsarkar  - we don't have a config example for Unsplash, so that "config example" link won't lead to anywhere. Would you like to add a config example as well? :) 

Already added please check. Thanks

PR #640, opened 2021-06-01
useNonce android support

### LINKED ISSUE #638: useNonce not passed to Android
## Issue

I'm working on a RN app that needs to connect to an inhouse built (it is still WIP) auth server. The backend team does follow strictly the OIDC protocols, so I have no problem connecting to it, however, as it is still being worked on, issues with it occur. For example, it's missing a nonce when returning the payload, so validating the token fails. While that's good for me to send as feedback, I believe there is a setting to disable the nonce check altogether -- `useNonce`. However, if you check the implementation over [here](https://github.com/FormidableLabs/react-native-app-auth/blob/main/index.js#L200), only iOS is being sent that config flag, not Android. Is there a specific reason for that that I'm unaware? This sounds like a bug to me, but if there is something else I'm missing -- please let me know.

---

## Environment

* **Your Identity Provider**: `Custom inhouse`
* **Are you using Expo?**: No

SleipRecx:
Fixes #638 #645 

## Description

Version 0.8.1 of [AppAuth-Android](https://github.com/openid/AppAuth-Android/releases/tag/0.8.1 ) added support for disabling nonce verification for value sent in the Authentication Request, useNonce is currently only passed to `authorizeWithConfiguration` on iOS, this pull-request adds support for Android.


kadikraman (approved):
Fantastic! Thanks for adding this 🙌

kadikraman:
Thanks so much for adding this! Could you please also update the tests to reflect the new variable for Android.

kadikraman (approved):

kadikraman:
Released in [v6.4.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v6.4.0) 🎉

Issue #600, opened 2021-01-27
On some Android devices, the application crashes after passing the authentication procedure


mishannn:
## Issue

On some Android devices, the application crashes after passing the authentication procedure. Works correctly on most devices.

**Config:**

```typescript
const { AUTH_CLIENT_ID, AUTH_CLIENT_SECRET, AUTH_REALM, AUTH_URL, AUTH_APP_SCHEME } = Config;

export function getAuthConfig(): AuthConfiguration {
    return {
        issuer: `${AUTH_URL}/realms/${AUTH_REALM}`,
        clientId: AUTH_CLIENT_ID,
        clientSecret: AUTH_CLIENT_SECRET || undefined,
        redirectUrl: `${AUTH_APP_SCHEME}:/callback`,
        scopes: ['openid'],
    };
}
```

**Error:**

```
Caught exception
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()' on a null object reference
 at com.rnappauth.RNAppAuthModule.onActivityResult(RNAppAuthModule.java:398)
 at com.facebook.react.bridge.ReactContext.onActivityResult(ReactContext.java:308)
 at com.facebook.react.ReactInstanceManager.onActivityResult(ReactInstanceManager.java:758)
 at com.facebook.react.ReactDelegate.onActivityResult(ReactDelegate.java:90)
 at com.facebook.react.ReactActivityDelegate.onActivityResult(ReactActivityDelegate.java:112)
 at com.facebook.react.ReactActivity.onActivityResult(ReactActivity.java:68)
 at android.app.Activity.dispatchActivityResult(Activity.java:7454)
 at android.app.ActivityThread.deliverResults(ActivityThread.java:4375)
 at android.app.ActivityThread.handleSendResult(ActivityThread.java:4424)
 at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:49)
 at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
 at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
 at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1818)
 at android.os.Handler.dispatchMessage(Handler.java:106)
 at android.os.Looper.loop(Looper.java:193)
 at android.app.ActivityThread.main(ActivityThread.java:6762)
[...]

PR #641, opened 2021-06-01
Android - Changed ActivityResult requestCode to be a value different than 0

KrasimirGatevMentorMate:
Android - Changed ActivityResult requestCode to be a value different than 0

Possible fix for #600  

## Description

Changed ActivityResult requestCode to be a value different than 0, as this can cause some unforeseen issues in Andorid and is not a good practice in general.

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -394,7 +394,7 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
      */
     @Override
     public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
-        if (requestCode == 0) {
+        if (requestCode == 52) {
             if (data == null) {
                 if (promise != null) {
                     promise.reject("authentication_error", "Data intent is null" );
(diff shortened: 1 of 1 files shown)


kadikraman:
Thanks for the PR! Could you explain why passing in `0` as the request code is not a good practice? I couldn't find anything specific when Googling this.

KrasimirGatevMentorMate:
Hi, thank you for the answer. The requestCode value is used so that for a given Activity, in onActivityResult() can distinguish between the results of different requests that were started with startActivityForResult(). The value should be as unique as possible so as to avoid colissions. Althout there is only one startActivityForResult() called in the library code currently, activities using different libraries such as QR code scanners etc. can also call their own startActivityForResult() and the codes could get mixed especially when using 0 as a value.

This change is an attemp at handling issues #600 which is experienced by many users including in our project in Production. This is a very hard to reproduce issues, but leads to NPE crashes, which have high impact for users, so I would be grateful if your team could take a more closer look at it. Although it is a bit of a "long shot" in any case in my oppinion, having requestCode be a concrete value will be a net improvement to the onActivityResult() function. Thank you for your time and hard work.

kadikraman:
Released in [v6.4.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v6.4.0) 🎉

PR #654, opened 2021-07-21
Improve iOS error handling

### LINKED ISSUE #653: iOS OAuth callback on-error behaviour differs from Android
## Issue

I'm entering this issue because I already have a PR to fix it (in draft at the time of writing while I check on tests - see #654), and the PR template requires an issue number.

When handling OAuth callback URI errors with `error` and (optionally) `error_description` parameters, on Android we see a raised exception from React Native App Auth where `error.code` contains the value that was in the `error` query string parameter and `error.message` contains the value that was in the `error_description` query string parameter. This makes sense; we get to access the server's error information and choose how or if to present relevant related information to the user.

On iOS, the integration with AppAuth-iOS differs from the Android layer by composing `error` and `error_description` into a single string that's put into `error.message`, e.g. `server_error: Foo`. This is inconsistent and can be obstructive, especially if one "owns" both the app and the server-side implementation, with the server side presenting information that we, as end-to-end system creators and maintainers, _do_ intend to show to the user.

It's easy to fix since AppAuth-iOS includes a complete dictionary of query parameters in the `NSError` object's `userInfo`, allowing us to extract the relevant information.

## Environment

* **Your Identity Provider**: All
* **Platform that you're experiencing the issue on**: iOS
* **Are you using Expo?**: I don't know what that is, so probably - no!

pond:
Fixes #653

## Description

The way that callback URI query data is turned into error information by AppAuth-iOS, then subsequently turned into a JavaScript exception by React Native App Auth, causes a divergence in behaviour between Android and iOS. See #653 for details.

This PR takes the `userInfo` dictionary in the `NSError` object and, if it can find a relevant entry, sets the error `message` to match the URI callback query string `error_description` parameter without prefixing or other modification. The prefix of error code was unnecessary, since this value is already present via the JavaScript exception object's `code` attribute.

I note while here that the long implementation of `-getErrorCode:defaultCode` could probably be replaced by just taking the `userInfo` dictionary and reading `[OIDOAuthErrorResponseErrorKey][OIDOAuthErrorFieldError]`, but I don't know the code base well enough to do that (and it'd be another PR anyway).

For reference, see:

* Constants at https://github.com/openid/AppAuth-iOS/blob/master/Source/AppAuthCore/OIDError.m#L39
* Method definition at https://github.com/openid/AppAuth-iOS/blob/master/Source/AppAuthCore/OIDErrorUtilities.m#L63 and `userInfo` dictionary assignment at https://github.com/openid/AppAuth-iOS/blob/master/Source/AppAuthCore/OIDErrorUtilities.m#L78
* Error handling passing in parsed query data (https://github.com/openid/AppAuth-iOS/blob/master/Source/AppAuthCore/OIDAuthorizationService.m#L137) as a dictionary at https://github.com/openid/AppAuth-iOS/blob/master/Source/AppAuthCore/OIDAuthorizationService.m#L142.

## Steps to verify

* Using this branch and some demo app of your choosing, visit an OAuth provider which can redirect back to the callback URL with `error` and `error_description` in the query string and provoke that condition.
* Catch the exception raised by React Native App Auth and inspect the object, noting that the `message` field value now matches the `error_description` URI query parameter value without being prefixed by the OAuth error code in the `error` URI query parameter value.

## A note on testing

There appears to be nothing I can find in the `index.spec.js` file that tests round-trip auth functionality; all tests appear concerned with data types and direct calling interfaces to the library via fairly extensive mocking. Consequently, I've not been able to add tests for the changed behaviour as I'd have to implement anew an entire test system to do that kind of testing, and I've no idea how to approach that... `:-(`

### DIFF
--- a/ios/RNAppAuth.m
+++ b/ios/RNAppAuth.m
@@ -35,7 +35,7 @@ static NSUInteger const kStateSizeBytes = 32;
 static NSUInteger const kCodeVerifierBytes = 32;
 
 RCT_EXPORT_MODULE()
-    
+
 RCT_REMAP_METHOD(register,
                  issuer: (NSString *) issuer
                  redirectUrls: (NSArray *) redirectUrls
@@ -221,7 +221,7 @@ RCT_REMAP_METHOD(refresh,
   return [OIDTokenUtilities encodeBase64urlNoPadding:sha256Verifier];
 }
 
-    
+
 /*
  * Perform dynamic client registration with provided OIDServiceConfiguration
  */
@@ -239,7 +239,7 @@ RCT_REMAP_METHOD(refresh,
     for (NSString *urlString in redirectUrlStrings) {
         [redirectUrls addObject:[NSURL URLWithString:urlString]];
     }
-    
+
     OIDRegistrationRequest *request =
     [[OIDRegistrationRequest alloc] initWithConfiguration:configuration
                                              redirectURIs:redirectUrls
@@ -248,7 +248,7 @@ RCT_REMAP_METHOD(refresh,
                                               subjectType:subjectType
                                   tokenEndpointAuthMethod:tokenEndpointAuthMethod
                                      additionalParameters:additionalParameters];
-    
+
     [OIDAuthorizationService performRegistrationRequest:request
                                              completion:^(OIDRegistrationResponse *_Nullable response,
                                                           NSError *_Nullable error) {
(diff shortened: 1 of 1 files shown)


kadikraman (approved):
This is awesome, thanks for adding this, and for the detailed explanation of your thought process in the PR description 🙌

Issue #643, opened 2021-06-02
Use web browser instead of modal in-app


kiki-le-singe:
## Issue

Hi !

First, thanks for this project. It helps me a lot ! :)

Is it possible AppAuth directly opens the web brower (Safari, Chrome) instead of a modal in-app ?

Thanks for your help

---

## Environment

* **Your Identity Provider**: `Custom`
* **Platform that you're experiencing the issue on**: `both`
* **Are you using Expo?** `no`

tdekoning:
I was encountering the same issue, and have created a pull request to add these options: #655

Jay-A-McBee:
Same as https://github.com/FormidableLabs/react-native-app-auth/issues/588

carbonrobot:
Fixed in #655

Issue #622, opened 2021-04-01
Support Custom User-Agents on iOS


martin-shaw:
AppAuth-iOS supports custom user agents https://github.com/openid/AppAuth-iOS#custom-user-agents-ios-and-macos https://github.com/openid/AppAuth-iOS/issues/200

I'm planning to fork and add this feature to support the out of the box custom browsers in AppAuth-iOS plus other custom browsers. My requirement is to support https://www.vmware.com/uk/products/workspace-one/enterprise-mobile-browser.html

I have found this feature has been requested here https://github.com/FormidableLabs/react-native-app-auth/issues/471

My question is, if I go ahead and add this feature to my fork is it likely to be accepted as a feature back into the main project?

ULTiALAt:
I'd be interested in this feature being added

TonyFTannous:
I'd be interested in this feature being added, any news

PR #655, opened 2021-07-21
Adding support for specifying what browser needs to be used for authorization

### LINKED ISSUE #588: The ability to specify browser in authorize method
Hello!

Maybe I missed something in the documentation and select wrong issue pattern for my question.

In our application, we faced with the problem of transition to the site of another service, access to which is also configured in SSO.

For some reason, when initially logging into the mobile app, chrome is used instead of the default browser. But to follow the link, android uses the default browser. As a consequence, session and cookies are not shared between the chrome browser and the default browser. And the user goes to the site unauthorized.

As far as I understand, if for some reason the default browser cannot be used for authorization in authorize method, then chrome will be used.

Is there a way to find out which browser was used to initially log into the application?
Is there a way to configure the use of a specific browser for authorization so that I can use it to follow links to other resources through SSO?

Thanks!

tdekoning:
This fixes #588 and #643 

## Description

This pull request adds the ability to control what browser is used for the authorization process. This helps in complex flows where the authorization is going through a browser and has a sidestep to a third party app.

In our case we were encountering issues, because third party apps use the default browser for return urls, but the in-app browser (Such as `aswebauthenticationsession`) doesn't share cookies with Safari.
By providing these parameters, the developers can fine-tune the login flow for their needs.

**Question for the maintainers:**
I'm not an iOS developer, and am seeing one flaw in the user flow when using an external browser:
If the user cancels the flow by pressing the "back" navigation on the top left of the browser, the promise isn't rejected. Do you have any idea how i can fix this? On Android, the promise is rejected when returning (which is expected)

## Steps to verify

On iOS: 
 - Adding `iosCustomBrowser` with the value `safari` should open safari, instead of a browser modal. 
 - Without providing `iosCustomBrowser`, the behavior should remain the same as before this pull request

On Android:
 - Adding `androidAllowCustomBrowsers` with the value ['chrome'] should open chrome, instead of the in-app browser
 - Without providing `androidAllowCustomBrowsers`, the behavior should remain the same as before this pull request

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ with optional overrides.
 - **usePKCE** - (`boolean`) (default: true) optionally allows not sending the code_challenge parameter and skipping PKCE code verification, to support non-compliant providers.
 - **skipCodeExchange** - (`boolean`) (default: false) just return the authorization response, instead of automatically exchanging the authorization code. This is useful if this exchange needs to be done manually (not client-side)
 - **iosCustomBrowser** - (`string`) (default: undefined) _IOS_ override the used browser for authorization, used to open an external browser. If no value is provided, the `SFAuthenticationSession` or `SFSafariViewController` are used.
+- **androidAllowCustomBrowsers** - (`string[]`) (default: undefined) _ANDROID_ override the used browser for authorization. If no value is provided, all browsers are allowed.
 
 #### result
 
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ with optional overrides.
 - **useNonce** - (`boolean`) (default: true) optionally allows not sending the nonce parameter, to support non-compliant providers
 - **usePKCE** - (`boolean`) (default: true) optionally allows not sending the code_challenge parameter and skipping PKCE code verification, to support non-compliant providers.
 - **skipCodeExchange** - (`boolean`) (default: false) just return the authorization response, instead of automatically exchanging the authorization code. This is useful if this exchange needs to be done manually (not client-side)
+- **iosCustomBrowser** - (`string`) (default: undefined) _IOS_ override the used browser for authorization, used to open an external browser. If no value is provided, the `SFAuthenticationSession` or `SFSafariViewController` are used.
 
 #### result
 
@@ -331,7 +332,7 @@ Make `AppDelegate` conform to `RNAppAuthAuthorizationFlowManager` with the follo
 + @property(nonatomic, weak)id<RNAppAuthAuthorizationFlowManagerDelegate>authorizationFlowManagerDelegate;
 ```
 
-Add the following code to `AppDelegate.m` (to support iOS <= 10 and React Navigation deep linking)
+Add the following code to `AppDelegate.m` (to support iOS <= 10, React Navigation deep linking and overriding browser behavior in the authorization process)
 
 ```diff
 + - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<NSString *, id> *) options {
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -241,8 +241,8 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
             final Promise promise
     ) {
         this.parseHeaderMap(headers);
-        final ConnectionBuilder builder = createConnectionBuilder(dangerouslyAllowInsecureHttpRequests, this.authorizationRequestHeaders, connectionTimeoutMillis, androidAllowCustomBrowsers);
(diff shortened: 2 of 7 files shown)


kadikraman:
Hi @tdekoning 👋  - thanks so much for taking a shot at this feature!

I just [merged a PR](https://github.com/FormidableLabs/react-native-app-auth/pull/654/files) to do with improving error messages on iOS which is unfortunately causing you come conflicts now. These should be pretty straightforward to solve - just merge main into your branch, choose your code from the diff and replace `error localizedDescription` with `self getErrorMessage: error`

Regarding user-cancelled-flow not working on iOS: I'm seeing this too, but I also don't get the success redirect. Can you confirm you've seen a successful auth flow using the Example app using Safari? Or was it using your own app?

tdekoning:
@kadikraman I've resolved the merge conflict

martin-shaw:
This is similar to the feature I requested/proposed doing here https://github.com/FormidableLabs/react-native-app-auth/issues/622

It would be nice if you could provide your own user agent to support other use cases, MDM managed browsers for example, rather than be limited to a fix set of browsers and have to make a code change to add support for a custom browser.

NicolaiOksen:
Any progress on this? It would be nice to be able to specify the desired browser for handling authentication.

kadikraman:
Apologies, I've been away from this project. Unfortunately there's merge conflicts again, I'm sorry! @tdekoning are you still around to resolve them?

joelfsreis:
It would be amazing to have this feature available! Any news @tdekoning?

ColtonSinning:
@tdekoning 
this is good point. when I use your PR, ios is working well
but android has an issue

ColtonSinning:
image

README.md
```diff
@@ -135,6 +135,8 @@ with optional overrides.
 - **useNonce** - (`boolean`) (default: true) optionally allows not sending the nonce parameter, to support non-compliant providers
 - **usePKCE** - (`boolean`) (default: true) optionally allows not sending the code_challenge parameter and skipping PKCE code verification, to support non-compliant providers.
 - **skipCodeExchange** - (`boolean`) (default: false) just return the authorization response, instead of automatically exchanging the authorization code. This is useful if this exchange needs to be done manually (not client-side)
+- **iosCustomBrowser** - (`string`) (default: undefined) _IOS_ override the used browser for authorization, used to open an external browser. If no value is provided, the `SFAuthenticationSession` or `SFSafariViewController` are used.
+- **androidAllowCustomBrowsers** - (`string[]`) (default: undefined) _ANDROID_ override the used browser for authorization. If no value is provided, all browsers are allowed.
```
eliaslecomte, line 139: Why is it not androidCustomBrowsers like the ios name?

tdekoning:
@kadikraman I'm sorry for the delay, i had to focus on some other things. 
Is this pull request a feature that is still desired for the library?

kadikraman:
Hi @tdekoning - I understand completely! We'd still love to have this feature. If you do find time to resolve the conflicts, please @ me in the PR and I'll review it immediately.

README.md
tdekoning: Because things work differently on iOS vs Android. It is normal to be able to select a browser to complete a task on Android, while on iOS you want it to just work without interrupting the user.

android-imdad:
@tdekoning Yes please, I would very much like to have this feature

tdekoning:
@kadikraman I've fixed the merge conflicts.
However, i wasn't able to test it properly, since i no longer have a server set up with which i can authenticate. The example in this repository also seems to not fully work.
I did test opening the correct browsers when using the `authorize` function within the example-app and that opened the correct browser.

It would be awesome if you could do some quality checks to see if this fix actually works in a real life scenario.

kadikraman:
Thank you @tdekoning!

I see what you mean regarding: the example app. Looks like https://demo.identityserver.io/ is down (hopefully not permanently?) and the Auth0 example seems to have also changed since we configured it so it no longer redirects back like it used to. Fun. I'll fix the configuration first.

We'll test this on real device to be sure. Based on the code, I'm expecting that this will open the custom browser, but if the user doesn't have it installed, it will fall back to the default.

android-imdad:
Hi @kadikraman Any update on merging this ? It would be great to have this soon.

kadikraman:
Looks like the identity server demo was moved so I updated the example app to point to the new instance https://github.com/FormidableLabs/react-native-app-auth/pull/780

Merging this now 🚀

kadikraman:
Just a head up I'm not able to publish this yet - looks like the latest `main` doesn't work for me on Android (getting invalid token on the identity server demo). I don't believe it's this change, but I don't want to do a release until we've confirmed it's not caused by one of the recently merged PRs.

kadikraman:

Issue #479, opened 2020-03-02
[Android] prefetchConfiguration fails if warmAndPrefetchChrome not defined


aeirola:
## Issue

On Android the call to `prefetchConfiguration` fails if `warmAndPrefetchChrome` is not included in the auth configuration. The typescript types define it as an optional parameter, so either the types are wrong, or the call should succeed even though the parameter isn't defined.

---

## Environment

* **Your Identity Provider**: `Auth0`
* **Platform that you're experiencing the issue on**: `Android`
* **Are you using Expo?**: `No`

PR #701, opened 2022-01-21
add configurable timeout fn param (connectionTimeoutSeconds/connectionTimeoutMillis)

### LINKED ISSUE #236: Ability to configure timeout for network request
Hi there,

I was trying to search the docs and the issues for any kind of configuration or values for the network request timeouts, but could not find anything.

My use case is that when there is a ongoing refresh, I would like to block any additional refresh requests before I can be sure the timeout duration has ran out.

Please let me know if there is already a way of achieve this with the current library version.

Jay-A-McBee:
Fixes #236 #479

## Description

- adds a connectionTimeoutSeconds fn param to prefetchConfiguration, authorize and refresh
- adds connectionTimeoutSeconds property to AuthConfiguration type
- adds connectionTimeoutSeconds param validation and test cases
- converts value to milliseconds for Android platform
- adds logic to native methods to handle new fn param
- adds android specific test cases for time conversion (seconds -> milliseconds)

Random addition
- adds default value for `warmAndPrefetchChrome` fn param (fixes 479)

## Steps to verify

(not sure if this is the best way to go about it 🤷‍♂️ )

1. Manually sync the Example/node_modules/react-native-app-auth directory with the updated ios/android dirs, react-native-app-auth.podspec and index.js files.

`cp -r ios Example/node_modules/react-native-app-auth && cp -r android Example/node_modules/react-native-app-auth && cp index.js Example/node_modules/react-native-app-auth && cp react-native-app-auth.podspec Example/node_modules/react-native-app-auth`

2. change the `issuer` property of the identityserver config in Example/App.js to a url that will never resolve (ex. http://www.example:81.com)

3. If the url doesn't use https protocol, add `dangerouslyAllowInsecureHttpRequests: true`  to the argument object passed to the 'authorize' method in Example/App.js.

4. Run the Example app and attempt to authorize through identity server. The default timeout is 60 seconds for iOS and 15 seconds for Android. It should be apparent that the configured timeout of 5 seconds is being respected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ This is the result from the auth server:
 - **scopes** - ([`string`]) the scopes the user has agreed to be granted
 - **authorizationCode** - (`string`) the authorization code (only if `skipCodeExchange=true`)
 - **codeVerifier** - (`string`) the codeVerifier value used for the PKCE exchange (only if both `skipCodeExchange=true` and `usePKCE=true`)
+- **connectionTimeoutSeconds** - (`number`) configure the request timeout interval in seconds.
 
 ### `refresh`
 
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ This is the result from the auth server:
 - **scopes** - ([`string`]) the scopes the user has agreed to be granted
 - **authorizationCode** - (`string`) the authorization code (only if `skipCodeExchange=true`)
 - **codeVerifier** - (`string`) the codeVerifier value used for the PKCE exchange (only if both `skipCodeExchange=true` and `usePKCE=true`)
-- **connectionTimeoutSeconds** - (`number`) configure the request timeout interval in seconds.
+- **connectionTimeoutSeconds** - (`number`) configure the request timeout interval in seconds. This must be a positive number. The default values are 60 seconds on iOS and 15 seconds on Android.
 
 ### `refresh`
 
@@ -202,12 +202,12 @@ This method will logout a user, as per the [OpenID Connect RP Initiated Logout](
 import { logout } from 'react-native-app-auth';
 
 const config = {
-  issuer: '<YOUR_ISSUER_URL>'
+  issuer: '<YOUR_ISSUER_URL>',
 };
 
 const result = await logout(config, {
   idToken: '<ID_TOKEN>',
-  postLogoutRedirectUrl: '<POST_LOGOUT_URL>'
+  postLogoutRedirectUrl: '<POST_LOGOUT_URL>',
 });
 ```
 
@@ -277,14 +277,14 @@ are not distributed as part of the bridge.
 
 AppAuth supports three options for dependency management.
 
-1. **CocoaPods**
+1.  **CocoaPods**
 
     ```sh
     cd ios
     pod install
     ```
 
-2. **Carthage**
+2.  **Carthage**
 
     With [Carthage](https://github.com/Carthage/Carthage), add the following line to your `Cartfile`:
 
@@ -296,7 +296,7 @@ AppAuth supports three options for dependency management.
 
     Add a copy files build step for `AppAuth.framework`: open Build Phases on Xcode, add a new "Copy Files" phase, choose "Frameworks" as destination, add `AppAuth.framework` and ensure "Code Sign on Copy" is checked.
 
-3. **Static Library**
+3.  **Static Library**
 
     You can also use [AppAuth-iOS](https://github.com/openid/AppAuth-iOS) as a static library. This
     requires linking the library and your project and including the headers. Suggested configuration:
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -95,6 +95,7 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
             final ReadableMap serviceConfiguration,
             final boolean dangerouslyAllowInsecureHttpRequests,
             final ReadableMap headers,
+            final Double connectionTimeoutMillis,
             final Promise promise
     ) {
         if (warmAndPrefetchChrome) {
@@ -102,7 +103,7 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
         }
 
         this.parseHeaderMap(headers);
-        final ConnectionBuilder builder = createConnectionBuilder(dangerouslyAllowInsecureHttpRequests, this.authorizationRequestHeaders);
+        final ConnectionBuilder builder = createConnectionBuilder(dangerouslyAllowInsecureHttpRequests, this.authorizationRequestHeaders, connectionTimeoutMillis);
         final CountDownLatch fetchConfigurationLatch = new CountDownLatch(1);
 
         if(!isPrefetched) {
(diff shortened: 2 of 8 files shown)


index.js
```diff
@@ -69,6 +69,13 @@ const validateAdditionalHeaders = headers => {
   );
 };
 
+const MILLI_PER_SEC = 1000;
+
+const convertTimeoutForPlatform = (platform, connectionTimeout) =>
+  platform === 'android' && connectionTimeout > 0
+    ? connectionTimeout * MILLI_PER_SEC
+    : connectionTimeout;
+
```
Jay-A-McBee, line 78: Of course iOS uses seconds and Android milliseconds 🤦 .

index.js
```diff
@@ -78,6 +85,7 @@ export const prefetchConfiguration = async ({
   serviceConfiguration,
   dangerouslyAllowInsecureHttpRequests = false,
   customHeaders,
+  connectionTimeoutSeconds = 0,
 }) => {
```
Jay-A-McBee, line 89: I was unable to pass a default value of null here because of the Java type. Even though Double can hold a `null` value, the conversion by React-Native of this arg to the Java type Double failed for null.

android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -63,6 +63,7 @@
     private final ReactApplicationContext reactContext;
     private Promise promise;
     private boolean dangerouslyAllowInsecureHttpRequests;
+    private Double connectionTimeoutMillis;
     private Boolean skipCodeExchange;
```
Jay-A-McBee, line 67: Used a Double instead of Integer because I saw this in the docs
Screen Shot 2022-01-20 at 7 20 03 PM

index.spec.js
```diff
@@ -42,6 +42,7 @@ describe('AppAuth', () => {
     usePKCE: true,
```
Jay-A-McBee, line 42: Some of the test cases seem off in this file. The `refresh` block has all the validation tests for `authorize` 🙀  - probably just a bad copy/pasta. Anyway, I'll get those shaped up.

index.spec.js
```diff
@@ -664,43 +780,52 @@ describe('AppAuth', () => {
...
+      require('react-native').Platform.OS = 'ios';
     });
 
+    const refreshToken = 'abc-token';
+
     it('throws an error when issuer is not a string and serviceConfiguration is not passed', () => {
       expect(() => {
-        authorize({ ...config, issuer: () => ({}) });
+        refresh({ ...config, issuer: () => ({}) }, { refreshToken });
       }).toThrow('Config error: you must provide either an issuer or a service endpoints');
```
Jay-A-McBee, line 791: Redundant `authorize` assertions that were mentioned.

README.md
```diff
@@ -151,6 +151,7 @@ This is the result from the auth server:
 - **scopes** - ([`string`]) the scopes the user has agreed to be granted
 - **authorizationCode** - (`string`) the authorization code (only if `skipCodeExchange=true`)
 - **codeVerifier** - (`string`) the codeVerifier value used for the PKCE exchange (only if both `skipCodeExchange=true` and `usePKCE=true`)
+- **connectionTimeoutSeconds** - (`number`) configure the request timeout interval in seconds.
```
kadikraman, line 154: Would be worth clarifying here that this needs to be a `positive and non-zero` (since looks like we're ignoring zeros). Also what's the default for both platforms if this isn't set.

android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -884,7 +886,7 @@ private AppAuthConfiguration createAppAuthConfiguration(
     /*
      *  Create appropriate connection builder based on provided settings
      */
-    private ConnectionBuilder createConnectionBuilder(boolean allowInsecureConnections, Map<String, String> headers) {
+    private ConnectionBuilder createConnectionBuilder(boolean allowInsecureConnections, Map<String, String> headers, Double connectionTimeoutMillis) {
```
kadikraman, line 889: The android build fails for me. Perhaps you forgot to copy over the code that adds the `connectionTimeoutMillis ` to createConnectionBuilder on the Android side?
```
23 actionable tasks: 2 executed, 21 up-to-date
/Users/kadi/Code/react-native-app-auth/Example/node_modules/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java:105: error: method createConnectionBuilder in class RNAppAuthModule cannot be applied to given types;
        final ConnectionBuilder builder = createConnectionBuilder(dangerouslyAllowInsecureHttpRequests, this.authorizationRequestHeaders);
                                          ^
  required: boolean,Map<String,String>,Double
  found: boolean,Map<String,String>
  reason: actual and formal argument lists differ in length
/Users/kadi/Code/react-native-app-auth/Example/node_modules/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java:164: error: method createConnectionBuilder in class RNAppAuthModule cannot be applied to given types;
        final ConnectionBuilder builder = createConnectionBuilder(dangerouslyAllowInsecureHttpRequests, this.registrationRequestHeaders);
                                          ^
  required: boolean,Map<String,String>,Double
  found: boolean,Map<String,String>
  reason: actual and formal argument lists differ in length
/Users/kadi/Code/react-native-app-auth/Example/node_modules/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java:237: error: method createConnectionBuilder in class RNAppAuthModule cannot be applied to given types;
        final ConnectionBuilder builder = createConnectionBuilder(dangerouslyAllowInsecureHttpRequests, this.authorizationRequestHeaders);
                                          ^
  required: boolean,Map<String,String>,Double
  found: boolean,Map<String,String>
  reason: actual and formal argument lists differ in length
/Users/kadi/Code/react-native-app-auth/Example/node_modules/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java:329: error: method createConnectionBuilder in class RNAppAuthModule cannot be applied to given types;
        final ConnectionBuilder builder = createConnectionBuilder(dangerouslyAllowInsecureHttpRequests, this.tokenRequestHeaders);
                                          ^
  required: boolean,Map<String,String>,Double
  found: boolean,Map<String,String>
  reason: actual and formal argument lists differ in length
/Users/kadi/Code/react-native-app-auth/Example/node_modules/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java:412: error: method createConnectionBuilder in class RNAppAuthModule cannot be applied to given types;
        final ConnectionBuilder builder = createConnectionBuilder(dangerouslyAllowInsecureHttpRequests, null);
                                          ^
  required: boolean,Map<String,String>,Double
  found: boolean,<null>
  reason: actual and formal argument lists differ in length
/Users/kadi/Code/react-native-app-auth/Example/node_modules/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java:507: error: method createConnectionBuilder in class RNAppAuthModule cannot be applied to given types;
                    createConnectionBuilder(this.dangerouslyAllowInsecureHttpRequests, this.tokenRequestHeaders),
                    ^
  required: boolean,Map<String,String>,Double
  found: boolean,Map<String,String>
  reason: actual and formal argument lists differ in length
6 errors
```

README.md
Jay-A-McBee: Ah it would probably make more sense to reuse the platform/native lib defaults here. 

AppAuth-Android [sets a default of 15 seconds](https://github.com/openid/AppAuth-Android/blob/master/library/java/net/openid/appauth/connectivity/DefaultConnectionBuilder.java#L38). AppAuth-iOS doesn't set this value so [the default is 60 seconds](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1408259-timeoutintervalforrequest/).

android/src/main/java/com/rnappauth/RNAppAuthModule.java
kadikraman: Ah, that makes sense!

kadikraman (approved):

ios/RNAppAuth.m
```diff
@@ -44,12 +44,13 @@ - (dispatch_queue_t)methodQueue
                  subjectType: (NSString *) subjectType
                  tokenEndpointAuthMethod: (NSString *) tokenEndpointAuthMethod
                  additionalParameters: (NSDictionary *_Nullable) additionalParameters
+                 connectionTimeoutSeconds: (double) connectionTimeoutSeconds
```
iAmAdamReid, line 47: This line appears to have introduced a regression: #790 

Proposed fix is here: #804

PR #705, opened 2022-02-04
Flip macCatalyst check for #704

### LINKED ISSUE #704: Latest release broken with macCatalyst
## Issue

It looks like #693 broke compilation for us on Catalyst. It blows up with OIDExternalUserAgentIOS is not available in macCatalyst.

The problem is TARGET_OS_IOS is set for Catalyst now per: https://stackoverflow.com/questions/12132933/preprocessor-macro-for-os-x-targets/49560690#49560690

Looks like simply flipping the if/elif order should fix things.

## Environment

* **Platform that you're experiencing the issue on**: `iOS (macCatalyst)`
* **Are you using Expo?** No

vincentjames501:
Fixes #704

## Description

See #704 for details

## Steps to verify

It compiles now on Catalyst :)


gispada:
Good catch, I actually didn't try to compile on Catalyst.

Jay-A-McBee (approved):

Issue #724, opened 2022-04-13
OAuth2 not working for dropbox


nmsdvid:
## Issue
OAuth2 is not working for dropbox, I'm using the example app and the example Dropbox config from the repo. The browser window opens and I'm able to login to Dropbox then it redirects to the redirectUrl but then nothing. The browser window is open and it shows the redirectUrl website. There is no callback in the app. Any ideas?

---

## Environment

* **Your Identity Provider**: `OAuth2 Dropbox`
* **Platform that you're experiencing the issue on**: `iOS`
* **Expo**: `no`
* **react-native**: `0.63.2` also on `0.68.0`
* **react**: `16.13.1` also on `17.0.2`
* **react-native-app-auth**: `^6.4.3`

Jay-A-McBee:
:wave: hello hello @nmsdvid - hmm... I think that our dropbox config example is probably out of date. It was committed 3+ years ago.  I was able to use Dropbox as an OAuth provider in the example app with a simple config.

```
{
  issuer: '',
  clientId: <CLIENT_ID>,
  redirectUrl: 'com.myapp.auth://oauth',
  scopes: [],
  serviceConfiguration: {
    authorizationEndpoint: 'https://www.dropbox.com/oauth2/authorize',
    tokenEndpoint: 'https://www.dropbox.com/oauth2/token',
  },
}
```

![dropbox](https://user-images.githubusercontent.com/13589602/163892334-b9e204a1-63fc-442a-8df3-031b2848b6bc.gif)

It seems like Dropbox now allows custom uri schemes as a `redirectUrl`. You shouldn't need to  create an https universal link or host an `apple-app-site-association` manifest on the associated domain anymore.

Some of the info in our example config is still accurate though. Dropbox does not expose a discovery endpoint (.well-known/openid-configuration) so a `serviceConfiguration` is necessary. It also handles a `refresh` token differently meaning you will not be able to refresh the user's auth token with this library's `refresh` method.

If you are able to get the dropbox API working, would you be willing to submit a PR that updates our outdated Dropbox config example?

Let me know how this works for you 🚀

nmsdvid:
Hello @Jay-A-McBee thanks for your help, it works. 😊  It seems Dropbox now needs custom uri scheme as a `redirectUrl`. I will test further the api and if everything works I will submit a PR, but for now I will close this ticket.

PR #727, opened 2022-04-26
Updated dropbox config example

nmsdvid:
Updated Dropbox config example

## Description

The current Dropbox config example is slightly outdated. I have updated the example to reflect the current working state.

## Steps to verify

If you would create a project with the current Dropbox config example you would have issues as it's described in this issue ticket [724](https://github.com/FormidableLabs/react-native-app-auth/issues/724). With the changes in the PR the Dropbox config example should work.


kadikraman (approved):
Awesome, thanks for the PR!

PR #736, opened 2022-06-08
Custom Nonce passing as an additional parameter is not properly configured

AnassHmida:
## Description

The nonce is not setup correctly once it's passed from JS to the native side as an additional parameter , resulting in a nonce comparison error.

When the nonce is passed as a parameter in the additional parameters section , and setting UseNonce to true , the native side does not take in consideration those changes and keeps the randomly generated nonce instead for comparison.

This PR helps checking if there is a nonce being passed in the additional parameters section and changes it if there is.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -56,6 +56,6 @@ repositories {
 dependencies {
     //noinspection GradleDynamicVersion
     implementation 'com.facebook.react:react-native:+'  // From node_modules
-    implementation 'net.openid:appauth:0.9.1'
+    implementation 'net.openid:appauth:0.11.1'
     implementation 'androidx.browser:browser:1.4.0'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -56,6 +56,6 @@ repositories {
 dependencies {
     //noinspection GradleDynamicVersion
     implementation 'com.facebook.react:react-native:+'  // From node_modules
-    implementation 'net.openid:appauth:0.11.1'
+    implementation 'net.openid:appauth:0.9.1'
     implementation 'androidx.browser:browser:1.4.0'
 }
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -56,6 +56,6 @@ repositories {
 dependencies {
     //noinspection GradleDynamicVersion
     implementation 'com.facebook.react:react-native:+'  // From node_modules
-    implementation 'net.openid:appauth:0.11.1'
+    implementation 'net.openid:appauth:0.9.1'
```
kadikraman, line 59: Any particular reason for downgrading the appauth library? It would be preferable to keep it at `0.11.1`

AnassHmida: @kadikraman Hello ! the downgrade is there because it's causing a conflict with the latest version of @react-native-firebase/perf which was giving me an " R8 RNAppAuthModule$5 is defined multiple times " error while running minify  using Hermes on release.

The error might be too specific , so i guess the version downgrade can be disregarded.

kadikraman: Hm yes I'm not sure what would cause that. You can always use [patch-package](https://github.com/ds300/patch-package) to change the version for your project till a solution can be found.

android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -674,9 +674,15 @@ private void authorizeWithConfiguration(
                 authRequestBuilder.setState(additionalParametersMap.get("state"));
                 additionalParametersMap.remove("state");
             }
+
+            if (additionalParametersMap.containsKey("nonce")) {
+                authRequestBuilder.setNonce(additionalParametersMap.get("nonce"));
+
```
kadikraman, line 680: Could we please also add
```java
additionalParametersMap.remove("nonce");
```
here? This will ensure it doesn't also get (redundantly) passed through as a url argument

AnassHmida: Hey @kadikraman , sure thing !

kadikraman (approved):

kadikraman:

PR #737, opened 2022-06-08
Pass additional param ui_locales

### LINKED ISSUE #730: [Android][Azure AD B2C] Unable to set additional parameter ui_locales
## Issue

Hello, I use this package for a react-native application (iOS & Android targets) backed by an Azure AD B2C tenant.
Most of the features work fine, but I facing an issue when trying to customise the language.
According to the documentation, a queryString parameter called `ui_locales` can be added to the login page in order to translate it.

I am trying to set it using `additionalParameters` as following:

```ts
// my login function, among other things, calling:
const result = await authorize({
  // required config with issuer, clientId, redirectUrl, scopes, serviceConfiguration etc.
  ...config,
  additionalParameters: {
    prompt: 'login',
    ui_locales: 'en',
  },
});
```

---

Now this works fine on iOS.
On Android however, the authentication is rejected (even before the redirection to AD login page):

> Error: Parameter ui_locales is directly supported via the authorization request builder, use the builder method instead

I found that [issue of AppAuth-Android](https://github.com/openid/AppAuth-Android/issues/726) referencing the problem.
A fix has [apparently been released in v0.10](https://github.com/openid/AppAuth-Android/issues/726#issuecomment-898654272), but as far as I know, the [latest release of this package uses v0.8.1](https://github.com/FormidableLabs/react-native-app-auth/blob/v6.4.3/android/build.gradle#L59).

Perhaps upgrading the android version will just solve the issue? 🤷 

## Environment

* **Your Identity Provider**: `Azure AD B2C`
* **Are you using Expo?** No

stbhb:
Fixes #730 - error "Parameter ui_locales is directly supported via the authorization request builder, use the builder method instead" when using ui_locales as additional parameter on Android.

## Description

Uses setUiLocales builder method when ui_locales was given with additional parameters.

## Steps to verify

1. Add ui_locales to additionalParameters in AuthConfiguration
2. Launch webbrowser for login by calling authorize
3. Error is thrown and the web browser never launches

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -674,6 +674,10 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
                 authRequestBuilder.setState(additionalParametersMap.get("state"));
                 additionalParametersMap.remove("state");
             }
+            if (additionalParametersMap.containsKey("ui_locales")) {
+                authRequestBuilder.setUiLocales(additionalParametersMap.get("ui_locales"));
+                additionalParametersMap.remove("ui_locales");
+            }
 
             authRequestBuilder.setAdditionalParameters(additionalParametersMap);
         }


kadikraman:

Issue #672, opened 2021-09-22
react-native-app-auth module getting crashed in Android


BajrangS:
## Issue

Firebase crashlytics report below crash issue in react-native app Android. There is no issues in iOS.

Caused by java.lang.NullPointerException
Attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()' on a null object reference

com.rnappauth.RNAppAuthModule.onActivityResult (RNAppAuthModule.java:414)

---

## Environment

* **Your Identity Provider**: `e.g. IdentityServer 4 / Okta / Azure`
Unable to identify because this is NPM module level crash

* **Platform that you're experiencing the issue on**: `iOS / Android / both`
Platform: Android

* **Are you using Expo?**
No

Please provide suggestions to fix this issue.

justinairbase:
Here's the crash log from my app
```

java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()' on a null object reference
    at com.rnappauth.RNAppAuthModule.onActivityResult(RNAppAuthModule.java:419)
    at com.facebook.react.bridge.ReactContext.onActivityResult(ReactContext.java:305)
    at com.facebook.react.ReactInstanceManager.onActivityResult(ReactInstanceManager.java:762)
    at com.facebook.react.ReactDelegate.onActivityResult(ReactDelegate.java:90)
    at com.facebook.react.ReactActivityDelegate.onActivityResult(ReactActivityDelegate.java:112)
    at com.facebook.react.ReactActivity.onActivityResult(ReactActivity.java:68)
    at android.app.Activity.dispatchActivityResult(Activity.java:8304)
    at android.app.ActivityThread.deliverResults(ActivityThread.java:5006)
    at android.app.ActivityThread.handleSendResult(ActivityThread.java:5054)
    at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:51)
    at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
    at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
    at android.os.Handler.dispatchMessage(Handler.java:106)
    at android.os.Looper.loop(Looper.java:223)
    at android.app.ActivityThread.main(ActivityThread.java:7664)
[...]

Issue #360, opened 2019-07-15
Unable to navigate away from Auth Component when Android OS destroys Activity


AlteaDown:
## Issue
The library can't seem to handle cases where `Activity.onDestroy()` is called when authenticating. Authentication is performed in a WebView separate from the Android Activity that is running RN, so that means when we move to it, the RN Activity is put on the backstack, and Android is now free to have the `Activity.onDestroy()` called on the Activity. 

When `Activity.onDestroy()` is called on the RN Activity, and once authentication succeeds, the library attempts to launch the previous activity again (but it was destroyed so it needs to be recreated). So we end up with a new Activity/Component that is completely unrelated to the authorize call we originally made. 

Some time after the new activity is done loading, our original call to `authorize()` will get a response from the token endpoint. However, the Component is detached, and unable to handle rendering. This means that if you want to navigate to a new Component, such as the rest of your app, you can not. The only way I can imagine being able to handle this is, first being aware that this situation can arise, when it does arise, and then to actually know that the state has changed on the new Component, you would have to use a reactive structure like a RxJS Subject that will emit to the new Component to allow it to handle this case, when it occurs.

You can recreate this scenario easily by using the "Don't Keep Activities" setting in your android Device's Developer Settings, or by using a low performance/low memory Android device.

Example code
```typescript
import React from 'react'
import {Button, Text} from 'native-base'
import {Image, ImageStyle, StyleSheet, View, ViewStyle} from 'react-native'
import {NavigationParams, NavigationScreenProps} from 'react-navigation'
import {authorize} from 'react-native-app-auth'
import Config from 'react-native-config'

type State = {
  showLogin: Boolean,
}

export default class LoginComponent extends React.Component<NavigationScreenProps, State> {
  constructor(props: Readonly<NavigationScreenProps<NavigationParams, State>>) {
    super(props)

    this.state = {
      showLogin: true,
    }
  }

  async authWithBackend() {
    this.setState({
      showLogin: false,
    })

    // use the client to make the auth request and receive the authState
    try {
      const authorizationResponse: any = await authorize({
        clientId: 'f3d259ddd3ed8ff3843839b',
        clientSecret: '4c7f6f8fa93d59c45502c0ae8c4a95b',
        redirectUrl: 'io.viamo.clipboard://',
        scopes: [],
        serviceConfiguration: {
          authorizationEndpoint: Config.OAUTH_ISSUER_URL,
[...]

Issue #564, opened 2020-09-30
[Android] additionalExchangeParameters cannot be null


sarahjaoun:
## Issue

Once I upgraded to the 5.1.2 version of the library, I started getting this crash on Android which wasn't the case before.
Firebase Crashlytics is showing that some android users are getting this crash and I am unable to replicate it.

Fatal Exception: java.lang.RuntimeException
Failure delivering result ResultInfo{who=null, request=0, result=-1, data=Intent { dat=com.saicomembers.portal:/callback?code=42c3e7fd5e18f675781670544e6cb7785d0e9e51ccac355a32f2d724476dc64f&scope=openid profile MobileApi offline_access&state=iiVkPbyNcTUftf_WkEWNgQ&session_state=aPQ4ArXSgr71BLZMM4YX_vX38tiI1qgRvlJ55Xe5pkM.b3709100ebfbecb216cd454f2d33b64f (has extras) }} to activity {com.saicohealth.members/com.saicohealth.members.MainActivity}: java.lang.NullPointerException: additionalExchangeParameters cannot be null

Below is the full stack trace:
net.openid.appauth.Preconditions.checkNotNull (Preconditions.java:55)
net.openid.appauth.AuthorizationResponse.createTokenExchangeRequest (AuthorizationResponse.java:454)
com.rnappauth.RNAppAuthModule.onActivityResult (RNAppAuthModule.java:401)
com.facebook.react.bridge.ReactContext.onActivityResult (ReactContext.java:262)
com.facebook.react.ReactInstanceManager.onActivityResult (ReactInstanceManager.java:703)
com.facebook.react.ReactActivityDelegate.onActivityResult (ReactActivityDelegate.java:124)
com.facebook.react.ReactActivity.onActivityResult (ReactActivity.java:75)
android.app.Activity.dispatchActivityResult (Activity.java:8393)
android.app.ActivityThread.deliverResults (ActivityThread.java:5442)
android.app.ActivityThread.handleSendResult (ActivityThread.java:5490)
android.app.servertransaction.ActivityResultItem.execute (ActivityResultItem.java:51)
android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:149)
android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:103)
android.app.ActivityThread$H.handleMessage (ActivityThread.java:2373)
android.os.Handler.dispatchMessage (Handler.java:107)
[...]

PR #743, opened 2022-06-23
Fix/android app crash on authorize 

nachoperez714:
Fixes https://github.com/FormidableLabs/react-native-app-auth/issues/600, https://github.com/FormidableLabs/react-native-app-auth/issues/672, https://github.com/FormidableLabs/react-native-app-auth/issues/360

Description
Fixes app crash issue on authorize for Android

Steps to verify
Enable "Don't Keep Activities" setting in your android Device's Developer Setting,
Authenticate user using authorize method

### DIFF
--- a/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -449,7 +449,7 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
 
             AuthorizationService authService = new AuthorizationService(this.reactContext, configuration);
 
-            TokenRequest tokenRequest = this.additionalParametersMap? response.createTokenExchangeRequest(this.additionalParametersMap) : response.createTokenExchangeRequest();
+            TokenRequest tokenRequest = this.additionalParametersMap != null? response.createTokenExchangeRequest(this.additionalParametersMap) : response.createTokenExchangeRequest();
 
             AuthorizationService.TokenResponseCallback tokenResponseCallback = new AuthorizationService.TokenResponseCallback() {
 
@@ -479,12 +479,12 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
             }
 
             SharedPreferences.Editor editor = sharedPref.edit();
-            editor.remove("dangerouslyAllowInsecureHttpRequests")
-            editor.remove("clientSecret")
-            editor.remove("dangerouslyAllowInsecureHttpRequests")
-            editor.remove("skipCodeExchange")
-            editor.remove("usePKCE")
-            editor.remove("codeVerifier")
+            editor.remove("dangerouslyAllowInsecureHttpRequests");
+            editor.remove("clientSecret");
+            editor.remove("dangerouslyAllowInsecureHttpRequests");
+            editor.remove("skipCodeExchange");
+            editor.remove("usePKCE");
+            editor.remove("codeVerifier");
             editor.apply();
 
         }
@@ -701,8 +701,8 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
         }
             SharedPreferences sharedPref = getCurrentActivity().getPreferences(Context.MODE_PRIVATE);
             SharedPreferences.Editor editor = sharedPref.edit();
-            editor.remove("dangerouslyAllowInsecureHttpRequests")
-            editor.apply()
+            editor.remove("dangerouslyAllowInsecureHttpRequests");
+            editor.apply();
     }
 
     private void parseHeaderMap (ReadableMap headerMap) {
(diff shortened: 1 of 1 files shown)


gksander:
I'm still working on testing this out and replicating the original issue. 

However, this line stands out to me:

```java
editor.putString("clientSecret", clientSecret);
```

This would be storing the client secret in an unencrypted text file, and that seems like perhaps a security vulnerability. It's recommended that you don't use a client secret in your app unless you _really have to_, so this perhaps isn't that much worse than just having the client secret in your app bundle in the first place. However, I'd like to think a bit about if there's a way to avoid storing such a sensitive piece of information unencrypted (or if this is really that much of a concern if you're already in the position of having the client secret in your app in the first place).

nachoperez714:
I appreciate that this could potentially be a security issue. Storing the value in the bundle, as you pointed out, is also insecure.

We could encrypted the value and then decrypt it later. However a vulnerability would still remain as the value would be accessible after use within active memory. This is because we are at the mercy of the JS engine garbage collection for when these objects stored in memory would be cleaned up once they are no longer in use. Per the MDN docs (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management) there is currently no way to explicitly trigger the garbage collection. Would it be possible to accept this pull request while we continue to investigate alternative methods to both fix the crash and improve security? At the moment this is a trade-off between a large number of our users experiencing crashes and a marginal security risk given the limitations of React Native.

nachoperez714:
@gksander 

Any new development on this?

gksander:
Sorry for the delay on this @nachoperez714. In terms of whether or not it's appropriate to merge this, I'm going to defer to @kadikraman and @Jay-A-McBee since they have much more knowledge than I do on this topic.

NullPainter2:
Hi guys, I don't have much experience with native Android, but docs says You can save data of Activity into a Bundle, and when it gets recreated You can load them from that Bundle  https://developer.android.com/guide/components/activities/activity-lifecycle#oncreate ... ???

kadikraman:
Hya! The reason for the indecision is that `SharedPreferences` is not a secure place to store data - any values stored in shared preferences will readable in plan text for any application or user with superuser privileges on a rooted device. Would this be a security risk for an auth library?

nachoperez714:
Have yoou considered implementing  @NullPainter2's suggestion to improve this fix? 

We are still getting crashes regarding this issue and would like to see it resolved.

dolly22:
Hello I was not able to replicate this exact issue with just setting "Don't Keep Activities" in developer settings, but we are also experiencing the crash for some of our users.

I have read all comments in linked issues and to my understanding the problem here probably is instance variables of RNAppAuthModule are set to initial state (it seems to me this is only possible when new instance is created through constructor). To my knowledge in ReactNative native module instance is created only once when application starts up.

@nachoperez714 Ale you able to catch this situation in debugger? What is creating new instance of RNAppAuthModule or settings those instance variables to their default values? Are you able to consistently replicate this issue somehow?

As suggested in #360 when the original activity is destroyed it leads to specific kind of problems, but in such case application is able to correctly resolve javascript promise and continue executing javascript code.

The null reference issues mentioned in #600 and #672 seem to me more like whole application process was already stopped and new one started when custom browser tab redirects back to application through net.openid.appauth.RedirectUriReceiverActivity and net.openid.appauth.AuthorizationManagementActivity (from https://github.com/openid/AppAuth-Android). If this would be the case (and I was not able to replicate it in emulator) I think there would be no easy fix due to RNAppAuth design (the Promise instance is already lost).

nachoperez714:
We weren't able to consistently reproduce the issue but we can still see it in AppDynamics impacting our users

image

SebastianHenn:
Hey team, I just wanted to check in on progress here as this is still an issue facing a number of our users. 

For me personally I can reproduce it consistently on a Samsung A13 with the following steps:
- Click on a firebase deep link to our app
- Install the app after being redirected to the playStore
- Launch the app for the first time
- Attempt to authorize, app crashes
- Subsequent launches of the app, either through deep link or regular do not cause the crash (in my instance, not sure if this is the case for our other users)

Interestingly the app also does not seem to crash if I install it regularly without having requested it via a deep link.

eithe:
It has now gone another year without updates from the maintainers. I know I'm getting this for free, but If you don't intend to keep supporting this library, please note that in the readme so people don't start using it.

This is a major showstopper for using this library. We are currently seeing this on 2% of sessions (according to Google crash analytics) which means that Google now has flagged the app for being above the "bad behavior threshold" resulting in the following: "Your 28-day user-perceived crash rate exceeds the bad behavior threshold of 1.09%. Your app is likely to be less discoverable on Google Play on all devices."

I have tried the patch that @yberstad suggested in https://github.com/FormidableLabs/react-native-app-auth/issues/564#issuecomment-1506812426 as a workaround, but haven't put it into production yet.

eithe:
Thanks! The patch I mentioned above (@yberstad in https://github.com/FormidableLabs/react-native-app-auth/issues/564#issuecomment-1506812426) has worked for us, will update here once I try this updated version, and get that to production.

VariabileAleatoria:
I think this issue needs attention, it's been two years since this pull request and we still do not have a solution

sanduluca:
Any updates to this ?

PR #747, opened 2022-07-13
Add support to TWA

tomferreira:
## Description

Add support to Trusted Web Activity on Android, a WebView-like experience for web content. This is useful for self-IdP use cases.

Its need to create the `assetlinks.json` file on the Authorization Server. See https://developer.chrome.com/docs/android/trusted-web-activity/ and  https://developers.google.com/digital-asset-links/v1/getting-started

If the service doesn't access the file or the device doesn't have Chrome installed, the behavior remains the same.


robwalkerco (approved):
Thanks for your PR!

robwalkerco:
@tomferreira I just wanted to let you know that this new flag has now been moved behind a new config value, so that it's opt-in. The change is released in [v7.1.0](https://github.com/FormidableLabs/react-native-app-auth/releases/tag/v7.1.0)

PR #749, opened 2022-07-13
Update Example Android build, link to local source

gksander:
Fixes #... 

## Description

This PR bumps the Android compile targets so that it can be installed on an Android emulator. Also uses local source code for the example (instead of using svn source).

## Steps to verify

- Clone the repo.
- Install example deps via `cd Example && yarn`.
- Run the Android example via `yarn Android`.

### DIFF
--- a/Example/App.js
+++ b/Example/App.js
@@ -1,8 +1,20 @@
-import React, { useState, useCallback, useMemo } from 'react';
-import { UIManager, Alert } from 'react-native';
-import { authorize, refresh, revoke, prefetchConfiguration } from 'react-native-app-auth';
-import { Page, Button, ButtonContainer, Form, FormLabel, FormValue, Heading } from './components';
-
+import React, {useState, useCallback, useMemo} from 'react';
+import {UIManager, Alert} from 'react-native';
+import {
+  authorize,
+  refresh,
+  revoke,
+  prefetchConfiguration,
+} from 'react-native-app-auth';
+import {
+  Page,
+  Button,
+  ButtonContainer,
+  Form,
+  FormLabel,
+  FormValue,
+  Heading,
+} from './components';
 
 const configs = {
   identityserver: {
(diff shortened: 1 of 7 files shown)


Example/App.js
```diff
@@ -1,8 +1,20 @@
-import React, { useState, useCallback, useMemo } from 'react';
-import { UIManager, Alert } from 'react-native';
-import { authorize, refresh, revoke, prefetchConfiguration } from 'react-native-app-auth';
-import { Page, Button, ButtonContainer, Form, FormLabel, FormValue, Heading } from './components';
-
+import React, {useState, useCallback, useMemo} from 'react';
```
gksander, line 1: Prettier took control here.

Example/App.js
```diff
@@ -48,7 +60,7 @@ const App = () => {
     prefetchConfiguration({
       warmAndPrefetchChrome: true,
       connectionTimeoutSeconds: 5,
-      ...configs.identityserver,
+      ...configs.auth0,
```
gksander, line 63: Note! the demo.identityserver has an expired cert, so this prefetch was actually failing and not surfacing the error to JS – it'd just hang on a white screen. Probably need to look into that at some point, but for now I just changed this to prefetch auth0 config.

Example/android/app/src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,8 @@
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:launchMode="singleTask"
-        android:windowSoftInputMode="adjustResize">
+        android:windowSoftInputMode="adjustResize"
+        android:exported="true">
```
gksander, line 19: `android:exported` was needed to run on newer Android API.

kadikraman (approved):
Awesome! 🚀

PR #868, opened 2023-06-27
Fixes InvalidTokenError 7.0.0-rc1 & 7.0.0-rc1 when using browser to authenticate

corne-de-bruin:
Fixes an InvalidTokenError that was introduced as part of MR #655 

## Description

Changed the _currentSession to use OIDAuthState in the skipCodeExchange false flow instead of OIDAuthorizationService
Also bumped the react-native version of the Example app to 0.68.7 as this was required to be able to run the Android app see: https://github.com/facebook/react-native/issues/35210

## Steps to verify

In the Example/App.js udpate the authorization config with:

```
iosCustomBrowser: 'safari',
androidAllowCustomBrowsers: ['chrome'],
iosPrefersEphemeralSession: false,
```

Login with safari on your iOS simulator and validate if it now returns an accessToken.

### DIFF
--- a/Example/android/build.gradle
+++ b/Example/android/build.gradle
@@ -3,7 +3,7 @@
 buildscript {
     ext {
         buildToolsVersion = "29.0.2"
-        minSdkVersion = 16
+        minSdkVersion = 21
         compileSdkVersion = 31
         targetSdkVersion = 31
     }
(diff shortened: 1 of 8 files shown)


robwalkerco (changes requested):
@corne-de-bruin I haven't been able to replicate this issue. Are you able to provide some additional replication details in the PR description that will allow us to validate the original issue and the fix?

Issue #861, opened 2023-06-13
Detect network errors on refresh token


bangerang:
Hi,

It appears that there's no robust way to check if the refresh request failed due to network errors. This is quite unfortunate, as it means you get logged out if you have no internet, which doesn't provide a great user experience. The only workaround I've found is to explicitly check if the error message contains "Connection error" for iOS and "Network error" for Android, but this feels far from ideal. Perhaps it would be possible to add another error code named "network_error" or something similar? Here's some code for reference:

``` javascript
try {
  await refreshToken();
  setIsLoggedIn(true);
} catch (error) {
  if (message.includes('Connection error') || message.includes('Network error')) {
    console.log('Connection error');
  } else {
    console.log('Refresh token failed, logging out...');
    setSessionExpired(true);
    setIsLoggedIn(false);
  }
}
```

bryceknz:
PR #886 (soon to be merged) should address the core issue here by giving apps a programmatic way to inspect the underlying native failure without matching against `error.message`.

It keeps the existing `error.message` / `error.code` behavior unchanged, but adds `error.nativeError` when the native SDK provides a lower-level cause, including refresh-token network failures.

Issue #799, opened 2023-01-09
error code token_exchange_failed.


moin-maniar:
## Issue

Getting network error after successfully login. This is my config.
const config = {
  issuer: 'http://localhost:9080/auth/realms/jhipster',
  clientId: 'web_app',
  redirectUrl: 'com.authproject.auth:/oauth2callback',
  scope: ['openid', 'nst-eid-oid', 'nst-email', 'nst-profile', 'nst-cloudid-authentication-method']
};

Please help me to resolve this issue.

---

## Environment
I am using [Keycloak](http://www.keycloak.org/)

PaulDardaine:
Hello, I don't know if you found the solution for your problem, but I got the same and I finally resolved it.

In your `AndroidManifest.xml` you should find this:
```
<application
      android:...
      android:networkSecurityConfig="@xml/network_security_config"
      android:...
```
If not, you can add this line.

In the `network_security_config` file (located at `app/src/main/res/xml`), ensure you have something like this
```
<?xml version="1.0" encoding="utf-8"?>
<network-security-config>
    <domain-config cleartextTrafficPermitted="true">
        <!-- If you are running on a device insert your computer IP -->
        <domain includeSubdomains="true">your.keycloak.server.ip</domain>

        <trust-anchors>
            <certificates src="system" />
            <certificates src="user" />
        </trust-anchors>
    </domain-config>

    <base-config cleartextTrafficPermitted="false" />
</network-security-config>
```

In my case, my keycloak server IP was the IP of my Mac because I'm running it locally for test purpose (192.168.X.X)

Hope this helps

emir0903d:
Hello, 
I am having the same error response after successful login to keycloak:
`{"nativeStackAndroid":[{"lineNumber":648,"file":"AuthorizationService.java","methodName":"doInBackground","class":"net.openid.appauth.AuthorizationService$TokenRequestTask"},{"lineNumber":579,"file":"AuthorizationService.j
ava","methodName":"doInBackground","class":"net.openid.appauth.AuthorizationService$TokenRequestTask"},{"lineNumber":394,"file":"AsyncTask.java","methodName":"call","class":"android.os.AsyncTask$3"},{"lineNumber":264,"file":"Fut
[...]

PR #886, opened 2023-08-23
Expose optional native error details for auth failures

KevinvdBurg:
## Description

Surfaces underlying native authorization errors on `error.nativeError` while keeping `error.message` unchanged (e.g. `Network error`).

Apps can keep showing `error.message` to users and use `error.nativeError` when they want the lower-level cause for debugging or custom UI.

This revives the original Android improvement, updates it for the current package layout, and extends the same behavior to iOS.

Related: #861, #799.

## Verification

- Verified Android locally with the demo app by forcing the IdentityServer token endpoint to use `https://expired.badssl.com`.
- Confirmed `error.message` stays `Network error` on both platforms.
- Confirmed the demo can consume `error.nativeError` and surface the certificate/cause detail (see screenshots below).
- Built and launched the iOS demo app, then verified the same broken token endpoint flow on iOS.

| Android | iOS |
| :-: | :-: |
| image | image |


robwalkerco:
Thanks for your PR @KevinvdBurg 

I noticed that you only modified the error description for Android. Is iOS already providing this additional error information? If not, would you be able to update your PR to keep iOS and Android aligned?

KevinvdBurg:
Hi @robwalkerco,

I'm trying to open the iOS project, but I'm getting the following error 🤔 
image

But also, I'm unsure where the iOS variant `handleAuthorizationException` is the naming is not consistent between the platforms. If you can point me to the right function, I can add it. :)

bryceknz:
Thanks for this contribution, @KevinvdBurg!

I was able to verify the Android behavior locally, and the added cause detail makes the error much more useful for debugging.

I’m going to continue the work from here by bringing the branch up to date with the current repo structure and extending the same underlying-error detail to iOS as well.

I'll make sure to give you git credit when I merge it!

bryceknz (approved):

PR #968, opened 2024-05-08
feat/detect network errors (android only) - BREAKING CHANGE

### LINKED ISSUE #861: Detect network errors on refresh token

faizplus:
Fixes #861 for Android only

## Description
This pull request adds error handling for network connectivity issues in the `refresh` and other function of this React Native module. Previously, the `refresh` or other function did not check for network connectivity before performing the token refresh, leading to potential issues when the device was not connected to a network. Introduced `network_error` error code for all functions. 

Though this is a useful feature, it will create a breaking change as we must follow the steps given below to use it.

### Allow permission to access the network state - BREAKING CHANGE
You need to add the `ACCESS_NETWORK_STATE` permission to your `AndroidManifest.xml` file. Here's how you can do it:

1. Open your `AndroidManifest.xml` file located in the `android/app/src/main` directory of your React Native project. 
2. Add the following line inside the `<manifest>` element, just before the `<application>` element:

```xml
<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
```
Your manifest should look something like this:

```xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="com.yourpackagename">

    <uses-permission android:name="android.permission.INTERNET"/>
    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>

    <application
        android:name=".MainApplication"
        android:label="@string/app_name"
        android:icon="@mipmap/ic_launcher"
        android:roundIcon="@mipmap/ic_launcher_round"
        android:allowBackup="false"
        android:theme="@style/AppTheme">
        ...
    </application>
</manifest>
```

## Steps to verify
1. Run the app
2. Authorize
3. Disconnect the Phone or Emulator from Internet
4. Try to use `refresh` function like below snippet.
5. It should have `network_error` in `error.code`
```javascript
 const handleRefresh = useCallback(async () => {
    try {
      const config = configs[authState.provider];
      const newAuthState = await refresh(config, {
        refreshToken: authState.refreshToken,
      });

      setAuthState(current => ({
        ...current,
        ...newAuthState,
        refreshToken: newAuthState.refreshToken || current.refreshToken,
      }));
    } catch (error) {
      Alert.alert('Failed to refresh token', error.code);
    }
  }, [authState]);
```


android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -429,6 +452,13 @@ public void logout(
                 dangerouslyAllowInsecureHttpRequests, androidAllowCustomBrowsers);
         final HashMap<String, String> additionalParametersMap = MapUtil.readableMapToHashMap(additionalParameters);
 
+        // Check if the device is connected to a network
+        if (!isNetworkConnected()) {
+            // If not connected, reject the promise with a network_error status and message
+            promise.reject("network_error", "No network connectivity");
+            return;
+        }
```
carbonrobot, line 460: I think you would need to move this after setting the promise on line 462, otherwise the return exits before setting the promise.

faizplus: yeah, let me update this one

robwalkerco:
> You need to add the `ACCESS_NETWORK_STATE` permission to your `AndroidManifest.xml` file. Here's how you can do it:

@faizplus I think it's possible to add the required permission to this library's own [Android Manifest file](android/src/main/AndroidManifest.xml), instead of requiring the user to update the apps Android Manifest manually. The library's and app's manifests are merged according to the behaviour described in the [Android Docs](https://developer.android.com/build/manage-manifests).

An example of this can be found in the Amplitude React Native SDK, where they specify the `ACCESS_COARSE_LOCATION` permissions in [their AndroidManifest file](https://github.com/amplitude/Amplitude-TypeScript/blob/01a8714f2cb2511959a84797ce95e8e35c612a9b/packages/analytics-react-native/android/src/main/AndroidManifest.xml#L3)

(Note: Although I believe this should work, I've never tried it myself)

faizplus:
@robwalkerco I tried adding `ACCESS_NETWORK_STATE` to this library's own [Android Manifest file](https://github.com/FormidableLabs/react-native-app-auth/pull/android/src/main/AndroidManifest.xml), and this is working perfectly fine. We can safely say that the user is not required to update the apps Android Manifest manually.

Let me know if this works for you, I will update this pr.

faizplus:
@carbonrobot I have updated the code to include `ACCESS_NETWORK_STATE` in the package itself so that the user don't have to add it manually as suggested by @robwalkerco. Now this PR is not introducing any breaking changes.

simoneb:
No expert here, but I thought I'd chime in as well because I'm also not sure if the permission is necessary, as suggested by @carbonrobot . I concur that intercepting a network error by parsing the error message is not great, but I'm not sure if adding a permission to check for a network error strikes a better balance. 

Unless there are any obvious reasons why parsing the error message in the library's code isn't viable option, I would go with that, and then expose to users a better way to intercept it, as you have already done in this PR.

PR #974, opened 2024-05-15
Fix/dupe state

### LINKED ISSUE #965: Sign in with Google on iOS with predefined state not possible
My **OIDAuthorizationRequest** needs a predefined **state**. Since there is no parameter in the interface I tried it with the additional parameters. That works but there is an issue with the default behaviour on **iOS** or in the objc implementation respectively:

RNAppAuth.m
```
    // builds authentication request
    OIDAuthorizationRequest *request =
    [[OIDAuthorizationRequest alloc] initWithConfiguration:configuration
                                                    clientId:clientId
                                            clientSecret:clientSecret
                                                     scope:[OIDScopeUtilities scopesWithArray:scopes]
                                               redirectURL:[NSURL URLWithString:redirectUrl]
                                              responseType:OIDResponseTypeCode
                                                     state: additionalParameters[@"state"] ? additionalParameters[@"state"] : [[self class] generateState]
                                                     nonce:nonce
                                              codeVerifier:codeVerifier
                                             codeChallenge:codeChallenge
                                      codeChallengeMethod: usePKCE ? OIDOAuthorizationRequestCodeChallengeMethodS256 : nil
                                      additionalParameters:additionalParameters];
```
Iff the state is present in the additional parameters, it gets taken from there but if it's there the lib will add it too. So the state is attached **twice**. 

For reference: The java code on the other hand removes the state from the additional parameters after reading the state. So it gets added only once. Which is the right approach if u ask me.

The Problem with the whole thing is, that the double attachment of the state leads to an **invalid_request** on the **google** side. (Apple seems not to be picky. At least if the key value pair is the same). looks like in the screenshot attached:
Screenshot 2024-04-29 at 08 40 41

My current workaround is; setting the state parameter to nil since the value is nullable and add it via the additional parameters from the react native js side.

---

## Environment
IDP: Zitadel / Google 
platform: iOS
react native: v0.73.4
react native app auth: ^7.1.0
No expo

### LINKED ISSUE #1114: Duplicate nonce if custom nonce is set on iOS
## Issue

My authConfig is something like so:
```
{
  useNonce: true,
  additionalParameters: {
    nonce: myCustomNonce
  }
}
```

When `authorize` is called, from Google I got: `OAuth 2 parameters can only have a single value: nonce`.

Looking into the source, on Android side, when `nonce` is set in `additionalParameters`, `nonce` is removed from `additionalParameters` after retrival: https://github.com/FormidableLabs/react-native-app-auth/blob/8fc717b07100a3f35382b122ed39ed5e0206537e/packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java#L703-L707

However, on iOS, we have https://github.com/FormidableLabs/react-native-app-auth/blob/8fc717b07100a3f35382b122ed39ed5e0206537e/packages/react-native-app-auth/ios/RNAppAuth.m#L340
And `nonce` is not removed from `additionalParameters`, whilst both `nonce` and `additionalParameters` are passed as parameters to `OIDAuthorizationRequest.initWithConfiguration`:https://github.com/FormidableLabs/react-native-app-auth/blob/8fc717b07100a3f35382b122ed39ed5e0206537e/packages/react-native-app-auth/ios/RNAppAuth.m#L343-L356 AFAIK, AppAuth-iOS will then put both the nonce twice in the URL: https://github.com/openid/AppAuth-iOS/blob/77e32a8bbfb973966692854c974c7599973a0f59/Sources/AppAuthCore/OIDAuthorizationRequest.m#L333-L365

The fix would be to align iOS with Android - `nonce` should be removed from `additionalParamters`.

---

## Environment

* **Your Identity Provider**: `Google`
* **Platform that you're experiencing the issue on**: `iOS`
* **Your `react-native` Version**: `0.83.6`
* **Your `react-native-app-auth` Version**: `8.1.0`
* **Are you using Expo?**: Yes

fant0mex:
Fixes #965
Fixes #1114

## Description

Removes `state` and `nonce` from iOS `additionalParameters` after reading them, so they are not sent twice to AppAuth-iOS.

This matches the existing Android behavior and fixes Google errors like:

> OAuth 2 parameters can only have a single value: state
> OAuth 2 parameters can only have a single value: nonce

## Verification

Manually verified in the iOS demo app using Google auth:

- `additionalParameters.state` completes without a duplicate `state` error.
- `additionalParameters.nonce` completes without a duplicate `nonce` error.


bryceknz:
Hey @fant0mex, long time no chat! I'm picking this PR up -- hope you don't mind. I'll make sure you get git credit 😊

I've rebased/merged it forward locally, kept your original `state` fix, and will add the matching `nonce` fix for #1114 since it has the same root cause.

bryceknz (approved):

PR #995, opened 2024-06-25
fix: iOS crash when scenes are enabled

kononenkoAnton:
Fixes #... 

## Description

iOS crashes when presenting a view controller while the application is using scenes: https://developer.apple.com/documentation/uikit/app_and_environment/scenes/
The issue occurs because each scene has its own window.
```
UIViewController *presentingViewController = appDelegate.window.rootViewController.view.window ? appDelegate.window.rootViewController : appDelegate.window.rootViewController.presentedViewController;
```
presentingViewController returns nil
IMG_7244

## Steps to verify

1. Prepare an application where scenes are enabled: https://developer.apple.com/documentation/uikit/app_and_environment/scenes/specifying_the_scenes_your_app_supports#3262273
2. On the RN side, call authorize from "react-native-app-auth"

### DIFF
--- a/.changeset/ten-needles-heal.md
+++ b/.changeset/ten-needles-heal.md
@@ -1,0 +1,5 @@
+---
+"react-native-app-auth": patch
+---
+
+fix: iOS crash when scenes are enabled
--- a/packages/react-native-app-auth/ios/RNAppAuth.m
+++ b/packages/react-native-app-auth/ios/RNAppAuth.m
@@ -130,7 +130,7 @@ RCT_REMAP_METHOD(authorize,
                                                                     return;
                                                                 }
                                                                 [self authorizeWithConfiguration: configuration
-                                                                 
+
                                                                                      redirectUrl: redirectUrl
                                                                                         clientId: clientId
                                                                                     clientSecret: clientSecret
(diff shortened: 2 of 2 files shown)


zibs:
Hey @kononenkoAnton thanks for the PR!

This does seem like it works, but do you think you could provide us with a really stripped down minimal example repo of a RN app that supports `scenes` as you mentioned.

I haven't used this before and have something clunky set up locally, but would love a reproducible app we could look at/test against in the future.

Let me know, thanks!

zibs:
Hey @kononenkoAnton - just wanted to give you a friendly ping to see if you'd be up for providing a repro here!

zibs:
Hey @kononenkoAnton I've been running the https://github.com/microsoft/react-native-test-app without your fix, and it seems like it works correctly still. The RNTA (react-native-test-app) has multiple window support enabled by default and seemed like a good place to test this.

I'm wondering if this might be an issue with your set up instead of an issue with the library.

Let me know if you think you can provide a stripped down example of your implementation, thanks!

sashrika:
Hi This PR really fix this issue. 
I manually added the following code to my node module and it started to work fine. 

 if (!presentingViewController) {
        presentingViewController = RCTPresentedViewController();
   }
    
   Please consider merging this. 
   Thank you.

AlixH:
Up ! Having the issue , the browser freezes/crashes when testing on iOS simulator

Issue #980, opened 2024-05-26
[iOS] RN 0.74 NewArch + bridgelesssMode ON authorize, refresh, logout doesn't work, Promise Rejection 


AptypTheKing:
## Issue

On iOS 17.4 RN 0.74.1 NewArch + **bridgelessMode ON(THIS IS IMPORTANT)** authorize, refresh, logout are throwing promise rejection. 

`Possible unhandled promise rejection: RNAppAuth.authorize(): Error while converting JavaScript argument 10 to objectiove C type BOOL. Objective C type BOOL is unsupported. 
`

![Screenshot 2024-05-26 at 20 19 33](https://github.com/FormidableLabs/react-native-app-auth/assets/106452723/677f6727-6744-4cdb-80dd-3fe0ab0bb0e1)
![Screenshot 2024-05-26 at 20 20 03](https://github.com/FormidableLabs/react-native-app-auth/assets/106452723/a177a9d2-60b9-4e2d-b4c6-138c06349871)

Here `useNonce = true` by default 

**ATTENTION: This happening only on bridgelessMode ON. Looks like some turbo modules issue.** 

---

## Environment
* **Your Identity Provider**: `Azure`
* **Platform that you're experiencing the issue on**: `iOS`
* **Your `react-native` Version**: `e.g. 0.74.1`
* **Your `react-native-app-auth` Version**: `e.g. 7.1.3`
* **Are you using Expo?** NO

carbonrobot:
We are actively working on Bridgeless support and hope to have an update soon.

steve4835:
I am having this issue as well

Johennes:
If anyone else is facing this, I used patch-package with

```
diff --git a/node_modules/react-native-app-auth/ios/RNAppAuth.m b/node_modules/react-native-app-auth/ios/RNAppAuth.m
index eb11aa5..938c045 100644
--- a/node_modules/react-native-app-auth/ios/RNAppAuth.m
+++ b/node_modules/react-native-app-auth/ios/RNAppAuth.m
@@ -97,10 +97,10 @@ - (dispatch_queue_t)methodQueue
                  skipCodeExchange: (BOOL) skipCodeExchange
                  connectionTimeoutSeconds: (double) connectionTimeoutSeconds
                  additionalHeaders: (NSDictionary *_Nullable) additionalHeaders
-                 useNonce: (BOOL *) useNonce
-                 usePKCE: (BOOL *) usePKCE
[...]

PR #1000, opened 2024-07-22
fix(ios): marshal booleans correctly (#999, #980)

### LINKED ISSUE #999: useNonce, usePCKE, and prefersEphemeralSession always false on iOS (and Mac Catalyst)
## Issue

On iOS (and Mac Catalyst), for the `authorize()` function, all boolean values passed from JS (namely `useNonce`, `usePCKE`, and `prefersEphemeralSession`) are marshalled incorrectly into Objective-C.

The method signatures all specify `(Bool *)` instead of `(Bool)` ([example](https://github.com/FormidableLabs/react-native-app-auth/blob/fe3646e1673e6fdb01d9ade0a36b5f59a514e860/packages/react-native-app-auth/ios/RNAppAuth.m#L100-L103)), so React Native marshals both `true` and `false` (from JS) to `NO` (Obj-C).

---

## Environment
* **Your Identity Provider**: Not sure, but affects all.
* **Platform that you're experiencing the issue on**: `iOS` (and `Mac Catalyst`)
* **Your `react-native` Version**: `0.73.1`
* **Your `react-native-app-auth` Version**: `7.2.0`
* **Are you using Expo?**: Yes

This was noticed in this issue: https://github.com/FormidableLabs/react-native-app-auth/issues/980#issuecomment-2223008235

shirakaba:
Fixes #999 and #980

## Description

Changes the method signatures from `(BOOL *)` (which marshals both `true` and `false` from JS to `NO` in Objective-C) to `(BOOL)` (which correctly marshals `true` to `YES` and `false` to `NO`).

## Steps to verify

1. Set a breakpoint inside the iOS `authorize` method.
2. Confirm that `useNonce = true` from JavaScript now becomes `YES` instead of `NO`.

### DIFF
--- a/.changeset/olive-terms-check.md
+++ b/.changeset/olive-terms-check.md
@@ -1,0 +1,5 @@
+---
+'react-native-app-auth': major
+---
+
+Breaking change (iOS, Mac Catalyst): The boolean values `useNonce`, `usePCKE`, and `prefersEphemeralSession` are now handled correctly. Previously, they were all being interpreted as `false` regardless of their actual values, but now the intended (`true` or `false`) value is correctly marshalled from JavaScript to native. To preserve behaviour from before this breaking change, explicitly set them all to `false`.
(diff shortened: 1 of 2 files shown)


sankar-gp:
May I know when we will get the update?

sankar-gp:
Could you let me know the approximate date for the publication of these changes and the release of the new version of the react-native-app-auth library?

zibs (approved):
Thanks a lot @shirakaba -- this looks good and seemed to work in my testing!

PR #1002, opened 2024-07-22
Feat/macos/support react native macos

### LINKED ISSUE #1001: react-native-macos support
I've got React Native macOS support working, so will try to prepare a PR for it!

shirakaba:
Fixes #1001

## Description

Adds [react-native-macos](https://github.com/microsoft/react-native-macos) support 🥳 

## Steps to verify

Follow the updated setup instructions in `examples/demo/README.md`.


examples/demo/metro.config.js
```diff
@@ -1,12 +1,32 @@
...
 );
 
+/**
+ * A crude flag to force Metro to resolve one or another fork of react-native.
+ * Set to "macos" to resolve react-native-macos.
+ * Set to "ios" to resolve react-native.
+ * Welcoming a fully automatic solution by anyone more familiar with Metro!
+ * @type {"macos" | "ios"}
+ */
+const applePlatform = "ios";
```
shirakaba, line 22: This is the best I could do with the time I have available. If anyone's more familiar with Metro, please show me how we can get Metro to resolve the correct one out of `react-native-macos` and `react-native` in this yarn monorepo.

examples/demo/package.json
```diff
@@ -5,13 +5,15 @@
...
     "ios": "react-native run-ios",
+    "macos": "react-native run-macos",
     "lint": "eslint .",
     "start": "react-native start",
     "test": "jest"
   },
   "dependencies": {
     "react": "18.2.0",
     "react-native": "0.72.4",
+    "react-native-macos": "~0.73.30",
```
shirakaba, line 16: This is the latest version of `react-native-macos`.

Although `react-native-macos` does provide iOS and Android implementations as well as macOS, I've avoided swapping `react-native` for `react-native-macos`, as I see there is ongoing work on New Architecture support and I don't want `react-native-macos` to block that with its lagging update cycle.

packages/react-native-app-auth/ios/RNAppAuth.m
```diff
@@ -499,35 +538,54 @@ - (void)endSessionWithConfiguration: (OIDServiceConfiguration *) configuration
...
+    }
+#endif
 
 #if TARGET_OS_MACCATALYST
     id<OIDExternalUserAgent> externalUserAgent = nil;
 #elif TARGET_OS_IOS
     id<OIDExternalUserAgent> externalUserAgent = iosCustomBrowser != nil ? [self getCustomBrowser: iosCustomBrowser] : [self getExternalUserAgentWithPresentingViewController:presentingViewController
                                                                                                                                     prefersEphemeralSession:prefersEphemeralSession];
+#elif TARGET_OS_OSX
+    id<OIDExternalUserAgent> externalUserAgent = nil;
```
shirakaba, line 574: I was unsure about this `externalUserAgent = nil`, but it does turn out that under-the-hood for the APIs we're calling, the AppAuth library makes an appropriate `OIDExternalUserAgentMac` for you so there's no need to make one yourself.

packages/react-native-app-auth/ios/RNAppAuth.m
```diff
@@ -499,35 +538,54 @@ - (void)endSessionWithConfiguration: (OIDServiceConfiguration *) configuration
...
                                              callback: ^(OIDEndSessionResponse *_Nullable response, NSError *_Nullable error) {
                                                           typeof(self) strongSelf = weakSelf;
                                                           strongSelf->_currentSession = nil;
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
                                                           [UIApplication.sharedApplication endBackgroundTask:rnAppAuthTaskId];
                                                           rnAppAuthTaskId = UIBackgroundTaskInvalid;
+#elif TARGET_OS_OSX
+                                                          // Brings this app to the foreground.
+                                                          [[NSRunningApplication currentApplication] activateWithOptions:(NSApplicationActivateAllWindows)];
+#endif
```
shirakaba, line 588: macOS AppKit apps don't get backgrounded, so we don't need to declare background tasks (but do benefit from refocusing the app).

packages/react-native-app-auth/react-native-app-auth.podspec
```diff
@@ -9,7 +9,8 @@ Pod::Spec.new do |s|
   s.license      = package['license']
   s.authors      = package['author']
   s.homepage     = package['homepage']
-  s.platform     = :ios, '10.0'
+  s.ios.deployment_target = '10.0'
+  s.osx.deployment_target = '10.15'
```
shirakaba, line 13: Technically, AppAuth [lets us](https://github.com/openid/AppAuth-iOS/blob/c89ed571ae140f8eb1142735e6e23d7bb8c34cb2/AppAuth.podspec#L34-L35) deploy down to the following versions:

```ruby
s.ios.deployment_target = "9.0"
s.osx.deployment_target = "10.12"
```

I've chosen `10.15` for this library just to align with the `react-native-macos` template's minimum deployment target.

packages/react-native-app-auth/ios/RNAppAuth.m
```diff
@@ -733,8 +791,13 @@ - (NSString*)getErrorMessage: (NSError*) error {
     }
 }
 
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
 - (id<OIDExternalUserAgent>)getExternalUserAgentWithPresentingViewController: (UIViewController *)presentingViewController
                                                      prefersEphemeralSession: (BOOL *) prefersEphemeralSession
+#elif TARGET_OS_OSX
+- (id<OIDExternalUserAgent>)getExternalUserAgentWithPresentingWindow: (NSWindow *)presentingWindow
+                                             prefersEphemeralSession: (BOOL *) prefersEphemeralSession
+#endif
```
shirakaba, line 800: Note that this `(BOOL *)` should actually be written as `(BOOL)`. We should change it to `(BOOL)` after merging my PR https://github.com/FormidableLabs/react-native-app-auth/pull/1000.

examples/demo/metro.config.js
Saadnajmi: Metro has some support for out of tree platforms, and https://github.com/microsoft/rnx-kit/tree/main/packages/metro-config has the correct defaults I think. It's what I have used for everywhere I added macOS support

examples/demo/package.json
Saadnajmi: I would keep the version of RN and RNM the same minor, so that they don't bring in conflicting transitive dependencies like metro, the CLI, etc..

packages/react-native-app-auth/ios/RNAppAuth.m
```diff
@@ -24,7 +24,9 @@ - (dispatch_queue_t)methodQueue
     return dispatch_get_main_queue();
 }
 
-UIBackgroundTaskIdentifier rnAppAuthTaskId;
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
```
Saadnajmi, line 27: I think Mac Catlyst is contained in the iOS ifdef. See the graph I have in this PRs' notes: https://github.com/facebook/react-native/pull/42278

Saadnajmi: And if this works on visionOS / tvOS, you could also TARGET_OS_IPHONE

shirakaba: Thanks for this great resource, I was unsure about it as well and just continued the existing conventions. I was actually tempted to do a `#if canImport(UIKit)` here, though not sure whether that's a Swift-only thing.

I'll sort these out once I'm free. It seems all of these could become `TARGET_OS_IOS` or in some cases the more broad `TARGET_OS_IPHONE` (where I think the commonality is that they all happen to include UIKit in their SDKs). The `AppAuth` CocoaPod that this wraps has [tvOS and watchOS](https://github.com/openid/AppAuth-iOS/blob/c89ed571ae140f8eb1142735e6e23d7bb8c34cb2/AppAuth.podspec#L36-L39) builds, both with reduced functionality, so it's definitely something to consider carefully.

shirakaba:
Thanks @Saadnajmi for your great guidance! I'll address those aspects once I'm free (think today I'll be working on other things, however).

shirakaba:
@Saadnajmi I've now migrated the PR to use react-native-test-app, which resolves my issues with the Metro bundler. I've also aligned to a common minor of React Native.

However, the demo app requires a custom AppDelegate (though doesn't seem to need anything custom on Android as far as I can see). How would I set that up?

Saadnajmi:
> @Saadnajmi I've now migrated the PR to use [react-native-test-app](https://github.com/microsoft/react-native-test-app), which resolves my issues with the Metro bundler. I've also aligned to a common minor of React Native, and have addressed all the `ifdef`s.
> 
> 
> 
> However, the demo app requires a custom AppDelegate to make it conform to `RNAppAuthAuthorizationFlowManager`. How would I set that up?

I don't think RNTA supports custom App Delegates, so that would be a feature request (@tido64 fyi). It _does_ support expo config plugins, does that let you modify the app delegate? https://github.com/microsoft/react-native-test-app/wiki/Config-Plugins

Though I'm realizing for macOS, that means you need 0.74 😅

shirakaba:
Unfortunately the one implementing Expo config plugins support for react-native-macos is none other than myself, so I can report that it’s not ready yet 😅

I’ve actually already completed the implementation of Expo config plugins, but need to rewrite the tests, which I was hoping to get at this evening. Still though, it would depend on Expo reviewing and merging it, and even then would only make it into SDK 51 (which expects RN 0.74 🥲) or later. So I think we’d either need a new feature in RNTestApp or go without it in this case.

shirakaba:
Okay, I've ripped out RNTA again and now have a mix of `react-native@0.72.4` and `react-native-macos@0.73.30`.

I've tried raising react-native to v0.73 to get them both on v0.73 and couldn't get iOS to build.

I've also tried dropping react-native-macos to v0.72 to get them both on v0.72 and couldn't get macOS to build.

This is the best I can do!

tido64:
> Would be able to get them both on the same version if RNTA supported custom AppDelegates, however.

I don't know if this helps, but I've had this branch for a while that I've just submitted: https://github.com/microsoft/react-native-test-app/pull/2160

shirakaba:
@tido64 Thanks so much for that PR! Sorry I wasn't able to review any closer – I've had my hands full this week. I'll try it out once I get back to this PR.

robwalkerco (commented):
Thanks for the great PR ❤️ 

I've spotted a few very minor issues that would be great if you could address before we merge.

shirakaba:
@robwalkerco Thank you for the review! Pending tasks in my mind:

- [ ] Adopt React Native Test App again (now that it supports custom AppDelegates)
- [ ] Update to React Native v0.74 (as it recently became available for React Native macOS)
- [ ] Use CocoaPods `1.15.2` as per review
- [ ] Add any setup instructions to cover macOS into `docs/docs/introduction.md` and `README.md`.

I've been a little burnt out lately so may not be able to jump straight on this but do hope to push this over the line.

.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+yarn-path ".yarn/releases/yarn-1.19.1.cjs"
```
carbonrobot, line 5: If we want to use a non-global version of Yarn, we should look at enabling corepack instead of checking in the binary. Since this is unrelated to this specific PR, it would be best to revert this change and open a separate PR for this specifically.

carbonrobot (changes requested):

.yarnrc
shirakaba: I was actually hoping to just install using global Yarn v1 here, but installation always failed with the following error:

> expected workspace package to exist for "eslint"

I looked up the error and it recommended this as the fix. If you know another way to get yarn to successfully install, I'd be happy to implement that.

shirakaba:
@tido64 I'm struggling to use `macos.withAppDelegate()` to set up the necessary custom AppDelegate. We need to do two things:

1. Make AppDelegate conform to `RNAppAuthAuthorizationFlowManager`. This is possible by referencing the following `examples/demo/config-plugin.js` via `"plugins": ["./config-plugin.js"]` in `examples/demo/app.json`:
  ```js
  const { createRunOncePlugin, } = require("@expo/config-plugins");
  const { mergeContents } = require("@expo/config-plugins/build/utils/generateCode");
  const { macos } = require("react-native-test-app/plugins/index");

  function withCustomAppDelegate(config) {
    return macos.withAppDelegate(config, (config) => {
      // See:
      // - examples/demo/node_modules/react-native-test-app/macos/ReactTestApp/AppDelegate.swift
      // - https://github.com/microsoft/react-native-test-app/pull/2160#issue-2435799831
      config.modResults.contents = mergeContents({
        tag: "config-plugin.js",
        src: config.modResults.contents,
        newSrc: "extension AppDelegate: RNAppAuthAuthorizationFlowManager {}",
        anchor: /extension AppDelegate {/,
        offset: 0,
        comment: "//",
      }).contents;

      return config;
    });
  }

  module.exports = createRunOncePlugin(
    withCustomAppDelegate,
    "test-plugin.js",
    "UNVERSIONED"
  );
  ```
  2. Express the following Obj-C as Swift in the AppDelegate:
  ```objc
  #import <React/RCTLinkingManager.h>
  
  // ... a few lines later:
  
  - (void)applicationWillFinishLaunching:(NSNotification *)notification {
    [[NSAppleEventManager sharedAppleEventManager] setEventHandler:self
                             andSelector:@selector(getURL:withReplyEvent:)
                           forEventClass:kInternetEventClass
                              andEventID:kAEGetURL];
  }
  
  - (void)getURL:(NSAppleEventDescriptor *)event withReplyEvent:(NSAppleEventDescriptor *)reply
  {
    NSString* urlString = [[event paramDescriptorForKeyword:keyDirectObject] stringValue];
    NSURL *url = [NSURL URLWithString:urlString];

    if ([self.authorizationFlowManagerDelegate resumeExternalUserAgentFlowWithURL:url]) {
      return;
    }
    
    [RCTLinkingManager getUrlEventHandler:event withReplyEvent:reply];
  }
  ```
  ... Unfortunately, the current AppDelegate has multiple occurences of the string "applicationWillFinishLaunching", so it is impossible to come up with a non-fragile RegExp pattern to use with it (particularly as the config plugin does not support multi-line regexes). The Expo Config AppDelegate Plugin moreover seems to no-op if there's more than one match.

Largely, this is down to the `mergeContents()` method from Expo Config Plugins being a very limited API to begin with. I suppose there's no need to use `mergeContents()`, actually – maybe we can just do a bit of manual string manipulation ourselves inside withAppDelegate 🤔

shirakaba:
@robwalkerco I'm proceeding to try to get this onto React Native Test App so that adding support for Windows in future will be simpler and to improve long-term maintainability (we'd be able to just bump the version of RNTA to upgrade all example apps). Would you be happy with migrating to RNTA as part of this PR or should it be a follow-up PR?

(Noting that I wasn't having much luck getting this working without RNTA – testing the react-native and react-native-macos example apps required manually flipping a variable in `metro.config.js`)

tido64:
> ... Unfortunately, the current AppDelegate has multiple occurences of the string "applicationWillFinishLaunching", so it is impossible to come up with a non-fragile RegExp pattern to use with it (particularly as the config plugin does not support multi-line regexes). The Expo Config AppDelegate Plugin moreover seems to no-op if there's more than one match.

I added some anchor points. I hope this helps: https://github.com/microsoft/react-native-test-app/pull/2188

.yarnrc
shirakaba: I managed to successfully install with global Yarn v1, so I've removed this.

shirakaba:
@robwalkerco I've:

- rebased onto `main`
- matched the minor version of `react-native` (v0.72) used in this repo
- removed RNTA
- removed the local `yarn`
- removed the need for a `applePlatform = "macos"` config flag in `metro.config.js`
- updated the `README.md`
- confirmed the app still works on iOS

image

image

Is there any blocker to merging this?

Issue #922, opened 2023-11-21
Microsoft Edge as Custom Browser on iOS


matice2:
Hi,

I need specifically the Microsoft Edge Browser to be used on iOS devices while authenticating. When trying to login in Safari, I get this message: You must use Microsoft Edge to access this resource. 

Has anyone already had this problem and solved it by using an alternative, please message me. 

Thank you!

carbonrobot:
Have you tried the `iosCustomBrowser` option in your configuration?

matice2:
@carbonrobot Hi, `iosCustomBrowser` does not have the option for the Edge browser. I have tried it with other browsers like Chrome or Firefox - it works for them, but not for Edge.

We have not solved this issue with the Edge. The conditional access for iOS devices has been adjusted so that Safari is allowed for the SSO. This way it works.

carbonrobot:
Ok, looks like we just need to add the option here and validate: https://github.com/FormidableLabs/react-native-app-auth/blob/aba51ba5e6d9d29d23aa2ae22e88d5e283da7fe6/index.d.ts#L82

faizplus:
@carbonrobot since we are using `AppAuth-iOS` and it does not have any method exposed for Microsoft Edge like  [CustomBrowserChrome](https://openid.github.io/AppAuth-iOS/docs/latest/interface_o_i_d_external_user_agent_i_o_s_custom_browser.html#a039552ad3c51ab61d4a039b642265a90) [CustomBrowserFirefox](https://openid.github.io/AppAuth-iOS/docs/latest/interface_o_i_d_external_user_agent_i_o_s_custom_browser.html#a3a1b5c23a72001c7f0152cc7c3024154)  [CustomBrowserOpera](https://openid.github.io/AppAuth-iOS/docs/latest/interface_o_i_d_external_user_agent_i_o_s_custom_browser.html#ac16ea4983c585e0436ddd5ab461daf1f) [CustomBrowserSafari](https://openid.github.io/AppAuth-iOS/docs/latest/interface_o_i_d_external_user_agent_i_o_s_custom_browser.html#a7d1c0649c388957b4283fce7f4fe8626), we will not be able to support Microsoft Edge browser for iOS.

https://github.com/FormidableLabs/react-native-app-auth/blob/aba51ba5e6d9d29d23aa2ae22e88d5e283da7fe6/ios/RNAppAuth.m#L697-L722

carbonrobot:
[...]

PR #1011, opened 2024-07-31
Edge browser support

macos107:
Fixes #... [](https://github.com/FormidableLabs/react-native-app-auth/issues/922)

## Description

Edge Browser Support

### DIFF
--- a/packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -1088,6 +1088,10 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
                     browserMatchers.add(VersionedBrowserMatcher.SAMSUNG_CUSTOM_TAB);
                     break;
                 }
+                case "edge": {
+                    browserMatchers.add(VersionedBrowserMatcher.EDGE_CUSTOM_TAB);
+                    break;
+                }
             }
         }
         return browserMatchers;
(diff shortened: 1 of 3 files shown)


packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -1088,6 +1088,10 @@ private BrowserMatcher getBrowserAllowList(ReadableArray androidAllowCustomBrows
                     browserMatchers.add(VersionedBrowserMatcher.SAMSUNG_CUSTOM_TAB);
                     break;
                 }
+                case "edge": {
+                    browserMatchers.add(VersionedBrowserMatcher.EDGE_CUSTOM_TAB);
```
m0hamad, line 1092: When I run `npx react-native run-android` I get the following:
```bash
error: cannot find symbol browserMatchers.add(VersionedBrowserMatcher.EDGE_CUSTOM_TAB);
```
As per these [docs](https://www.javadoc.io/doc/net.openid/appauth/latest/net/openid/appauth/browser/VersionedBrowserMatcher.html)

However, when I comment this out, and set my default browser to Edge in the Android Emulator, the Example app does successfully open the links in Edge.

packages/react-native-app-auth/ios/RNAppAuth.m
```diff
@@ -714,6 +714,10 @@ - (NSString*)getErrorCode: (NSError*) error defaultCode: (NSString *) defaultCod
         @"firefox":
             ^{
                 return [OIDExternalUserAgentIOSCustomBrowser CustomBrowserFirefox];
+            },
+        @"edge":
```
m0hamad, line 718: I get the following error when running `npx react-native run-ios` or when I try to build directly within Xcode: `No known class method for selector 'CustomBrowserEdge'`

Same here, I was able to confirm that I can open the URLs within the buttons in the Example app in Edge for iOS, but only after commenting this out.

packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java
zibs: Thanks yeah, I don't think Edge is actually an option in AppAuth?

PR #1045, opened 2024-12-13
Add claims parameter to authorize

frereit:
## Description

Currently, the `setClaims` method of App Auth is not exposed. By adding a new key `claims` to `AuthConfig`, similar to `scopes`, when it is set, we can call `setClaims` on the `AuthorizationRequest.Builder` to add the requested claims to the authorization request.

## Steps to verify

Create a `AuthConfiguration` that requests some claims about the user:

```typescript
const config: AuthConfiguration = {
  issuer: 'http://example',
  clientId: '....',
  redirectUrl: 'localhost:8080',
  scopes: ['openid', 'offline_access'],
  claims: { "id_token": { "email": null } }
}
```
Then, use `authorize` to obtain an idToken, and verify that the requested claims are included in the token.

## Draft

This PR implements the `claims` key in the `AuthConfiguration` on the Android side. However, I do not own iOS devices and can therefore not implement this for iOS. I am opening this PR in the hopes that someone else completes it for the iOS side.


packages/react-native-app-auth/index.d.ts
```diff
@@ -68,6 +68,7 @@ export type BaseAuthConfiguration = BaseConfiguration & {
 export type AuthConfiguration = BaseAuthConfiguration & {
   clientSecret?: string;
   scopes: string[];
+  claims: object,
```
carbonrobot, line 71: We should consider constraining this type further, perhaps something like

```
claims: Record<string, string | number>;
```

carbonrobot (commented):
Thanks for the PR! Although we don't technically support Open ID Connect (OIDC), I can't see any reason not support this.

- [ ] Add support in AppAuth-iOS
- [ ] Documentation updates

packages/react-native-app-auth/index.d.ts
frereit: Yes, I considered this while writing the PR, but it does not quite work.

The best I can come up with is:

```suggestion
  claims?: {
    id_token?: Record<string, object | null>,
    userinfo?: Record<string, object | null>,
  }
```

This is because a Claims request can become quite complex, as described in Section 5.5 of the OIDC spec. The value of the `claims` key is "a JSON object listing the requested Claims", and each claim may have some more properties associated with it. I believe my suggested type here is even too restrictive, as the ODIC spec allows for more top level members, which then should be implementation-defined:

> Other members MAY be present. Any members used that are not understood MUST be ignored. 

Here is an example for a (complex) Claims request from the spec:

```json
  {
   "userinfo":
    {
     "given_name": {"essential": true},
     "nickname": null,
     "email": {"essential": true},
     "email_verified": {"essential": true},
     "picture": null,
     "http://example.info/claims/groups": null
    },
   "id_token":
    {
     "auth_time": {"essential": true},
     "acr": {"values": ["urn:mace:incommon:iap:silver"] }
    }
  }
```
 
It's up to you to decide how we want to type this, but either way, we should  make the key optional at the very least, otherwise this becomes a breaking change and is it also obviously not a required key.

```suggestion
  claims?: object,
```

Thanks for taking a look at this PR!

carbonrobot: Great callout, this is good context we can preserve here. Similar to [the JSON type debate](https://github.com/microsoft/TypeScript/issues/1897) its not entirely possible to correctly model JSON, so the try...catch is the backstop.

carbonrobot: I'm good with making it optional and leaving it as `object`

PR #1064, opened 2025-03-12
fix(getCustomBrowser): explicitly set browser to nil if type not found

### LINKED ISSUE #1063: iosCustomBrowser falsely evaluated as provided, causing crash in release builds (iOS)
## Issue

When calling the authorize method in RNAppAuth to open a private (ephemeral) browser session, the app crashes in release builds. The problem appears to be caused by the iosCustomBrowser parameter being evaluated as non‑nil even though it is not passed from the JavaScript API. As a result, RNAppAuth calls getCustomBrowser with an invalid value, leading to an EXC_BAD_ACCESS (SIGSEGV) crash in RNAppAuth.m at line 720.

**Expected behavior:**
The app should open a private browser session (ephemeral session) by utilizing iosPrefersEphemeralSession: true and successfully complete the authentication flow.

**Actual behavior:**
The app crashes with an EXC_BAD_ACCESS error when building in release mode.

### Steps to Reproduce
1. Configure the authorize method with a configuration object that includes:
```
const config = {
  issuer: "https://youridprovider.com",
  clientId: "yourClientID",
  redirectUrl: "yourapp://callback",
  scopes: ["openid"],
  iosPrefersEphemeralSession: true,
  usePKCE: true,
};
```
2. Call authorize(config) in your React Native code.
3. Build the app in release mode (this issue is only reproducible in release builds for iOS).
4. Observe that the app crashes with an EXC_BAD_ACCESS error.

### Additional Information
- I have tried configuring the authorize method both with and without the iosCustomBrowser parameter; previously, it worked without this parameter with no issue.
- When using iosCustomBrowser set to 'safari', it opens a regular browser session, not ephemeral.
- When iosCustomBrowser is not set, the app crashes.
- This issue has been reported on GitHub (see issue #1061), but no pull request is available yet.

---

## Environment

* **Your Identity Provider**: `other`
* **Platform that you're experiencing the issue on**: `iOS (release builds only)`
* **Your `react-native` Version**: `0.78.0`
* **Your `react-native-app-auth` Version**: `8.0.1`
* **Are you using Expo?** `No`

valentinrey:
Fixes #1063 

## Description

This pull request fixes an issue in the `getCustomBrowser:` method where an invalid or empty browser type (e.g., an empty string or NSNull) received from the JavaScript layer would not be properly handled. In iOS release builds, such values caused the method to pass a non-nil result, leading to an attempt to execute an invalid browser block and resulting in a crash.

- Updated `getCustomBrowser:` in `RNAppAuth.m` to explicitly check the return value of `browser`. If the value is not in the dictionary, it will return `nil` to trigger an ephemeral session. 
- Updated gitignore.

## Steps to verify

1. Configure the authorize method with a configuration object with no `iosCustomBrowser`, that includes:
```
const config = {
  issuer: "https://youridprovider.com",
  clientId: "yourClientID",
  redirectUrl: "yourapp://callback",
  scopes: ["openid"],
  iosPrefersEphemeralSession: true,
  usePKCE: true,
};
```
2. Call authorize(config) in your React Native code.
3. Build the app in iOS release mode (the issue is only reproducible in release builds for iOS).
4. Observe that the app not crashes.

### DIFF
--- a/.changeset/nervous-donuts-relate.md
+++ b/.changeset/nervous-donuts-relate.md
@@ -1,0 +1,5 @@
+---
+'react-native-app-auth': patch
+---
+
+#1063 Updated getCustomBrowser: in RNAppAuth.m to explicitly check the return value of browser. If the value is not in the dictionary, it will return nil to trigger an ephemeral session.
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,10 @@ android/gradle/
 android/gradlew
 android/gradlew.bat
 
+# Android/Eclipse
+#
+android/.settings/
+
 # node.js
 #
 node_modules/
--- a/packages/react-native-app-auth/ios/RNAppAuth.m
+++ b/packages/react-native-app-auth/ios/RNAppAuth.m
@@ -717,6 +717,9 @@ RCT_REMAP_METHOD(logout,
             }
     };
     BrowserBlock browser = browsers[browserType];
+    if (!browser) {
+        return nil;
+    }
     return browser();
 }
 #endif


tuladharjaa:
@valentinrey logout does not work after these changes

valentinrey:
@tuladharjaa, I had to refactor my logout logic a bit. Basically moved the check up in the file tree

sepperousseau:
@valentinrey what do you mean with "moved the check up in the file tree"?
For me the login works now, but the logout doesn't work anymore.

valentinrey:
@tuladharjaa I have an application with a simple auth logic, and take a check in App.tsx now that was in another file. It checks if the user is authenticated, if not, runs a logout function that also wipes tokens and other stored data

seanfar:
@sepperousseau @tuladharjaa 

I've introduced what I think may be the fix for `logout` in #1070. Please take a look at your convenience to see if this addresses the issue you're experiencing. ty!

Issue #1061, opened 2025-02-19
The app crashes if iosCustomBrowser is not provided, react-native 0.77.1


mluksha:
## Issue

The release configuration of the app crashes on iOS if **authorize** method is called _without_ **iosCustomBrowser** option. It reproduces only with react-native v0.77.1.

```
- (id<OIDExternalUserAgent>)getCustomBrowser: (NSString *) browserType {
    typedef id<OIDExternalUserAgent> (^BrowserBlock)(void);
    
    NSDictionary *browsers = @{
        @"safari":
            ^{
                return [OIDExternalUserAgentIOSCustomBrowser CustomBrowserSafari];
            },
        @"chrome":
            ^{
                return [OIDExternalUserAgentIOSCustomBrowser CustomBrowserChrome];
            },
        @"opera":
            ^{
                return [OIDExternalUserAgentIOSCustomBrowser CustomBrowserOpera];
            },
        @"firefox":
            ^{
                return [OIDExternalUserAgentIOSCustomBrowser CustomBrowserFirefox];
            }
    };
    BrowserBlock browser = browsers[browserType];
    return browser();  <-- Thread 1: EXC_BAD_ACCESS (code=1, address=0x10)
}
```

---

## Environment

* **Your Identity Provider**: IdentityServer 4
* **Platform that you're experiencing the issue on**: iOS (only release configuration)
* **Your `react-native` Version**: 0.77.1
* **Your `react-native-app-auth` Version**: 8.0.0
* **Are you using Expo?**: 52.0.35

tuladharjaa:
I also am facing this issue. It works fine on iOS simulator

SunlightKnight:
I'm facing the same issue:

- `react-native`: 0.77.1
- `react-native-app-auth`: 8.0.1

Seems to be working fine while debugging (either on the simulator or on the device), but the release build crashes at the `return browser()` line.

ksm:
Hello folks. The app I work on also just hit this issue.

The snippet where the problem [happens is here](https://github.com/FormidableLabs/react-native-app-auth/blob/main/packages/react-native-app-auth/ios/RNAppAuth.m#L377). Specifically:
```
id<OIDExternalUserAgent> externalUserAgent = iosCustomBrowser != nil ? [self getCustomBrowser: iosCustomBrowser] : nil;
```

[...]

PR #1070, opened 2025-04-25
fix(iOS): Handle `NSNull` in `iosCustomBrowser` param for logout

seanfar:
## Description

While there isn't an issue created specifically for what this PR is attempting to address, there are a few issues that seem to be related to it.

#1061 for example, which led to the fix introduced by #1064 addressed an issue on the `authorize` function. 

Unfortunately, the `logout` function is still failing to properly display the browser, causing the flow to silently fail when `iosCustomBrowser` is not provided, as the call to `[self getCustomBrowser:NSNull]` returns `nil`. 

This seems to be happening on RN V0.77.1 with the new architecture enabled. Untested on other versions.

This PR adds a guard to fallback to the default external user agent when `iosCustomBrowser` is `nil` or `NSNull`.

## Steps to verify

1. Configure the configuration object with no value for the `iosCustomBrowser` option that includes:

```js
const config = {
  issuer: "https://youridprovider.com",
  clientId: "yourClientID",
  redirectUrl: "yourapp://callback",
  scopes: ["openid"],
  iosPrefersEphemeralSession: true,
  usePKCE: true,
};
```

2. Add the calls accordingly for `authorize(config)` and `logout(config)` in your code.
3. Build the app in **iOS release mode**.
4. Execute the `authorize` (login) and `logout` calls via your application.
5. Observe that the default browser is displayed properly.

### DIFF
--- a/packages/react-native-app-auth/ios/RNAppAuth.m
+++ b/packages/react-native-app-auth/ios/RNAppAuth.m
@@ -517,8 +517,13 @@ RCT_REMAP_METHOD(logout,
 #if TARGET_OS_MACCATALYST
     id<OIDExternalUserAgent> externalUserAgent = nil;
 #elif TARGET_OS_IOS
-    id<OIDExternalUserAgent> externalUserAgent = iosCustomBrowser != nil ? [self getCustomBrowser: iosCustomBrowser] : [self getExternalUserAgentWithPresentingViewController:presentingViewController
-                                                                                                                                    prefersEphemeralSession:prefersEphemeralSession];
+    id<OIDExternalUserAgent> externalUserAgent;
+    if (iosCustomBrowser != nil && ![iosCustomBrowser isEqual:[NSNull null]]) {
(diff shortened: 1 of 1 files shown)


sepperousseau:
@seanfar I tested it, it fixes the problem for me. Thanks! 👍

zibs:
Hey there @seanfar, I'm trying to reproduce the error locally but don't seem to be able too. Can you verify I'm following the right steps?

I am on a more updated RN version with new arch:

```
"react": "19.0.0",
"react-native": "0.79.2",
```

my config is;

```
    issuer: 'https://demo.duendesoftware.com',
    clientId: 'interactive.public',
    redirectUrl: 'io.identityserver.demo:/oauthredirect',
    additionalParameters: {},
    scopes: ['openid'] as const,
    iosPrefersEphemeralSession: true,
    usePKCE: true,
```

and I am running in release mode on iOS:


https://github.com/user-attachments/assets/ff6df3b9-498d-4a52-9be7-2b20ef3a6ec9

Issue #917, opened 2023-11-01
Support for response mode is missing


paulschreiber:
## Issue
Support for response mode is missing. If you try to add it as an additional parameter:

```js
import {authorize} from 'react-native-app-auth';

const appleConfig = {
  issuer: 'https://appleid.apple.com',
  clientId: APP_CONFIG.appleClientId,
  redirectUrl: APP_CONFIG.appleRedirectURI,
  scopes: ['name', 'email'],
  additionalParameters: {
    response_mode: "form_post",
  }
};
let result = await authorize(appleConfig);
```

[You get this error](https://github.com/openid/AppAuth-Android/blob/c6137b7db306d9c097c0d5763f3fb944cd0122d2/library/java/net/openid/appauth/AdditionalParamsProcessor.java#L63):
>Parameter response_mode is directly supported via the authorization request builder, use the builder method instead

Please add a `responseMode` parameter to config. In RNAppAuthModule.java, use `setResponseMode()` to pass that on. A similar change is needed for iOS.

This may also require an upstream fix in `net.openid.appauth.AuthorizationRequest` to support `form_post`.

## Related issues
Implement "form_post" response mode  https://github.com/FormidableLabs/react-native-app-auth/issues/580
Has anyone used this lib for Apple Sign In (on Android)? https://github.com/FormidableLabs/react-native-app-auth/issues/484

## Environment

* **Your Identity Provider**: `Apple`
* **Platform that you're experiencing the issue on**: `Android `
* **Your `react-native` Version**: `0.72.6`
* **Your `react-native-app-auth` Version**: `7..1.0`

paulschreiber:
Upstream: https://github.com/openid/AppAuth-Android/issues/608

PR #1113, opened 2026-04-09
fix(android): added support for response mode parameter

### LINKED ISSUE #1112: Error while using response_mode as an additional parameter (Android)
## Issue

I am using this package in a react-native application (Android) which uses Okta identity provider.
I am facing an issue during authentication when I pass response_mode, as an additional parameter.
_My IdP's default response mode is form_post. Hence, I need to explicitly set `response_mode: query` to get authentication process working for my app._

Below is the snippet:

```
const result = await authorize({
  // issuer, clientId, redirectUrl, scopes etc.
  additionalParameters: {
    response_mode: 'query',
  }
});
```

After initiating the authentication process, it fails before redirecting to our Okta login page.

> Error: Parameter response_mode is directly supported via the authorization request builder, use the builder method instead

I believe, we will have to set it in `RNAppAuthModule.java` using `setResponseMode()` to pass it forward.

---

## Related issues

Support for response mode is missing #917

---

## Environment

* **Your Identity Provider**: `Okta (organization)`
* **Your `react-native` Version**: `0.72.6`
* **Your `react-native-app-auth` Version**: `7.1.0`

thesanjayrathod:
Fixes #1112

## Description

Handled response_mode parameter in RNAppAuthModule.java

## Steps to verify

Using `response_mode` parameter in authorization request as below, should not result into an error.

```
const result = await authorize({
  // issuer, clientId, redirectUrl, scopes etc.
  additionalParameters: {
    response_mode: 'query',
  }
});
```

### DIFF
--- a/.changeset/khaki-forks-unite.md
+++ b/.changeset/khaki-forks-unite.md
@@ -1,0 +1,5 @@
+---
+'react-native-app-auth': minor
+---
+
+Added support for passing response_mode as an additional parameter in the config.
--- a/packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java
+++ b/packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java
@@ -710,6 +710,10 @@ public class RNAppAuthModule extends ReactContextBaseJavaModule implements Activ
                 additionalParametersMap.remove("ui_locales");
 
             }
+            if (additionalParametersMap.containsKey("response_mode")) {
+                authRequestBuilder.setResponseMode(additionalParametersMap.get("response_mode"));
+                additionalParametersMap.remove("response_mode");
+            }
 
             authRequestBuilder.setAdditionalParameters(additionalParametersMap);
         }


bryceknz:
Hi @thesanjayrathod,

Thanks for the fix! I tested this locally on Android with `response_mode: 'query'` and confirmed it fixes the error from #1112.

I also tested `response_mode: 'form_post'` using the demo IdentityServer config. The previous Android builder error did not occur; the flow advanced and failed later with `Response state param did not match request state`.

So this appears to fix the Android-side rejection of `response_mode`, which is the shared failure in #1112 and #917. However: should #917 still stay open? It also mentions:
- the Apple Sign In `form_post` case
- possible upstream AppAuth-Android limitations
- iOS support
- maybe a first-class `responseMode` config option

I could not verify the Apple case end-to-end because I do not have an Apple Developer account.

firstwhitewalker:
Hi @bryceknz - yep, this fix is only for Android. So yes, as #917 also mentions about IOS, it should stay open.
This will only close #1112 

Thank you for testing the fix!

bryceknz (approved):

Issue #932, opened 2023-12-21
When executes authorize function it get stuck after gets open in browser with no further procedure with no error and no response


RitulPatil:
Getting stuck once open in web with no response

I have implemented as follows:
 const config:any = {
    issuer: '************',
    clientId: '*********',
    redirectUrl: '************',
    scopes: ['openid'],
    clientAuthMethod :'basic' ,
    
    // iosPrefersEphemeralSession: true,
    // androidTrustedWebActivity: true
  };
and using it in function
const azureAdLogin = async () => {

    try {
      const result = await authorize(config);
      console.log('azureAdLogin result=', result);
    } catch (error: any) {
      console.log("error=", error);
    }
  }

---

## Environment

 experiencing the issue on:   both`
react-native` Version: ` 0.72.7`
react-native-app-auth` Version: `7.1.0`

endrits079:
+1

janroures:
+1

Did you find any solution?

endrits079:
My issue was in the backend, I don't know tho what the backend guys did to fix, I am just assuming the redirect urls were not right

carbonrobot:
Thanks for the issue report, we are going to need some additional details in order to help out. Do you have any specific debugging information you can provide?

LekZ974:
+1
Everything works fine when logging in for the first time, but the browser freezes when "authorize" is triggered upon reconnection without kill the application.

sirmong:
I can log in, log out and log in again. But if I delete an account, then the browser freezes on a new log in attempt. Relaunch application helps.

No errors received

bogrees:
`react-native-app-auth` authorize call is freezing in iOS 18.
RN: 0.75.3
react-native-app-auth: 8.0.0

fabiohvp:
I'm also having this issue, can't even log in.. whenever I call authorize safari opens and freeze with a white screen (it doesn't even render azure page).

gothraven:
+1, anyone got a fix for this ?

I actually hooked up the `Safari Technology Preview` to inspect the content, it says that the webpage exists, it's just not rendered for some reasons cause the web browser pop up froze i guess

For context:
ReactNative: v0.63.5
ReactNativeAppAuth: v8.0.1 (also tried v8.0.0)

gothraven:
[...]

PR #1122, opened 2026-06-05
feat(ios): natively intercept https redirect URIs via ASWebAuthenticationSession callback (iOS 17.4+)

### LINKED ISSUE #987: [iOS] Universal link not working as redirect URI
## Issue

Hi, 

I am facing the problem that my https redirect URI is not working when using the normal authentication through the browser popup window. However, if I set the customBrowser to Safari, it works. This will open up the Safari browser in a external window. After swichting back to the popup window it works, but only for a limited time. After some time the Universal links stop working (Properly because the associated domains are cached somehow?). I have the feeling the apple app site association file is not loaded correctly, when the external browser is not opened once.  

The use of the external (or default web browser) variant is not accepted by Apple App Review anymore, because of bad user experience.

If anyone could solve this problem, please contact me!

Thanks!
---

## Environment
Keycloak
* **Platform that you're experiencing the issue on**: `iOS`
* **Your `react-native` Version**: `e.g. 0.72.1`
* **Your `react-native-app-auth` Version**: `e.g. 7.0.0`
* **Are you using Expo?**
no

danchily2:
Fixes #987

## Description

When the redirect URI is an https universal link, `OIDExternalUserAgentIOS` creates the session with `callbackURLScheme:@"https"` — a scheme `ASWebAuthenticationSession` does not support — so the session never intercepts the redirect. Flow completion then depends on the universal link opening the app from *inside* the auth session, which is not triggered by server-side redirects or JS navigation, and is sporadically dropped entirely. The result is an `authorize()` promise that never settles, with the user stuck on a permanently disabled login UI (also reported as #932; root cause tracked upstream since 2018 in openid/AppAuth-iOS#367, and the `callbackURLScheme:` initializer is deprecated per openid/AppAuth-iOS#847). The same failure class affects other SDKs (e.g. google/GoogleSignIn-iOS#388). The common workaround — a hosted "trampoline" page forwarding the callback to a custom scheme — reintroduces the custom scheme universal links were adopted to avoid, and the in-session JS hop is itself sporadically blocked by WebKit.

iOS 17.4 added `ASWebAuthenticationSessionCallback callbackWithHTTPSHost:path:`, which lets the session intercept an https redirect natively — no app-site-association timing dependency, no trampoline page, no custom scheme.

This PR adds `RNAppAuthHTTPSExternalUserAgent` (modeled on AppAuth's `OIDExternalUserAgentIOS`, including ephemeral-session and presentation-context handling) and uses it automatically when:

- no `iosCustomBrowser` is requested, and
- the redirect URI scheme is `https`, and
- running on iOS 17.4+.

Everything else — custom browsers, custom-scheme redirect URIs, iOS < 17.4 — keeps the existing behavior, so this is a progressive enhancement with no API surface change.

Notes per CONTRIBUTING:
- iOS-only by design: Android already supports https App Links redirects natively through AppAuth-Android's `RedirectUriReceiverActivity` intent filters, so there is no equivalent defect to replicate there.
- No JS/TypeScript surface change → no `index.js`/`index.spec.js`/typings/readme updates needed; `yarn test` and `yarn lint` are unaffected (native-only change).
- Changeset included (`minor`).

## Steps to verify

**Requirement:** the https callback path needs the callback host registered as an associated domain with the **webcredentials** service type — both the `com.apple.developer.associated-domains` entitlement (`webcredentials:example.com`) and a `webcredentials` entry for the app in the domain's `apple-app-site-association` file. Without the association, the agent transparently falls back to the legacy `callbackURLScheme` session (AppAuth's default behavior), so apps without the association are unaffected.

1. Configure an OIDC provider with an **https universal-link** redirect URI, add the `webcredentials` associated domain (entitlement + AASA), and run on an iOS 17.4+ device or simulator.
2. Call `authorize()` and complete sign-in: the session intercepts the redirect itself — the sheet closes on the provider's 302 without loading the redirect page, and `authorize()` resolves with the token response.
3. Cancel the sheet: `authorize()` rejects with the user-cancelled error as before.
4. Remove the `webcredentials` association and repeat: the fallback path engages and behavior matches current `main`.
5. Run on iOS < 17.4 (or with `iosCustomBrowser`, or a custom-scheme redirect URI): behavior is unchanged from current `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/https-callback-ios.md
+++ b/.changeset/https-callback-ios.md
@@ -1,0 +1,5 @@
+---
+'react-native-app-auth': minor
+---
+
+iOS: natively intercept https (universal link) redirect URIs on iOS 17.4+ using the ASWebAuthenticationSession https callback, so the authorization flow no longer depends on universal-link activation from inside the auth session — which is not triggered by server redirects and sporadically leaves authorize() pending forever (#987, #932).
--- a/.changeset/https-callback-ios.md
+++ b/.changeset/https-callback-ios.md
@@ -1,5 +1,0 @@
----
-'react-native-app-auth': minor
----
-
-iOS: natively intercept https (universal link) redirect URIs on iOS 17.4+ using the ASWebAuthenticationSession https callback, so the authorization flow no longer depends on universal-link activation from inside the auth session — which is not triggered by server redirects and sporadically leaves authorize() pending forever (#987, #932).
--- a/.changeset/https-callback-ios.md
+++ b/.changeset/https-callback-ios.md
@@ -2,4 +2,4 @@
 'react-native-app-auth': minor
 ---
 
-iOS: natively intercept https (universal link) redirect URIs on iOS 17.4+ using the ASWebAuthenticationSession https callback, so the authorization flow no longer depends on universal-link activation from inside the auth session (which is not triggered by server redirects and sporadically leaves authorize() pending forever).
+iOS: natively intercept https (universal link) redirect URIs on iOS 17.4+ using the ASWebAuthenticationSession https callback, so the authorization flow no longer depends on universal-link activation from inside the auth session — which is not triggered by server redirects and sporadically leaves authorize() pending forever (#987, #932).
--- a/.changeset/https-callback-ios.md
+++ b/.changeset/https-callback-ios.md
@@ -1,0 +1,5 @@
+---
+'react-native-app-auth': minor
+---
+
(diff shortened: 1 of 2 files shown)


shidoro:
@danchily2 is this PR a temporary solution until openid foundation adds support on their side, or is it intended to diverge from the underlying dependency, and FormidableLabs will maintain it?

What are the differences between the external user agent you're introducing, and the one from the library? I'm asking because I use react-native-app-auth, and a patched version of the external user agent coming from this PR https://github.com/openid/AppAuth-iOS/pull/938

If I'm understanding your PR correctly, it's going to intercept my https callback, and route it to your custom agent. Am I going to notice any difference?

PR #1128, opened 2026-08-28
fix(android): isolate auth request state and make prefetch asynchronous

OskarEichler:
## Fixes

Snapshot token headers, TLS policy, timeout, parameters, client authentication, PKCE verifier and promise per interactive flow. Keep refresh/registration independent, reject overlapping browser flows without replacing the first, and settle late token failures on their originating promise. Replace the blocking/global prefetch latch with per-issuer asynchronous completion; expose native prefetch completion and errors through the JS promise.

## Breaking / observable changes

Observable changes: omitted Android header groups no longer reuse previous headers; supply required headers on every call. A second pending authorize/logout rejects with authentication_in_progress. Await/catch prefetchConfiguration, which now settles after native work instead of immediately; iOS prefetch remains a no-op. Browser cancellation and discovery failure release the slot, while already-started exchanges keep their own state. Existing AuthorizationService/CustomTabs disposal work in #1127 is not duplicated. This is an Android module-state fix, not a fix for the separate iOS process-global NSURLSession limitation.

## Verification

14 compiled actual-method Android diagnostics pass (11 fail on the baseline), plus four JS prefetch controls. Tests include two providers, omitted header groups, pending authorize plus refresh, delayed token failure, cancellation, cache reuse, per-issuer discovery and nonblocking errors. A fresh read-only boundary investigation and a separate candidate review found no remaining Android module-state mixing route. Network/browser/bridge dependencies are controlled doubles; no live-provider or physical-device claim.

Each patch was applied independently to upstream commit `6f9090ce1b991c0acc2a6f5dc8bea838deaf4f7a` and passed its targeted external actual-source diagnostics or standalone type/lockfile check. The combined fixes pass 75 focused diagnostics. These use synthetic values, controlled native/bridge doubles and disposable filesystem projects, not real credentials.

Combined branch checks:

- Existing Jest suite: 3 suites, 81 tests pass.
- Root lint, demo lint, targeted App.tsx lint, demo TypeScript and standalone public-type checks pass.
- Expo plugin TypeScript build passes. The optional lint:plugin command has a baseline tooling limitation: ESLint 6 does not find its TypeScript-only input.
- Android Debug demo build passes with RN 0.79.2.
- Unsigned iOS Simulator Debug demo build passes with RN 0.79.2.
- Android and iOS production Metro demo bundles pass.
- git diff --check passes.

No checked-in test/spec files were added, changed or disabled. The source changes are supported by external reproducible diagnostics because this audit's repository policy prohibits checked-in test edits. Live OAuth providers, physical-device/browser authentication and a full Expo native build were not exercised. React Doctor reports two warnings: the discovery response is explicitly guarded by response.ok === false, and the demo pending reset is inside finally; both paths are exercised by focused checks. No rules were suppressed.

## Scope

- `packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java`
- `docs/docs/usage/prefetch.md`
- `packages/react-native-app-auth/index.js`
- `packages/react-native-app-auth/index.d.ts`
- `docs/docs/usage/config.md`
- `.changeset/appauth-android-request-lifecycle.md`

Unrelated audit fixes are submitted separately. The separate iOS process-global OIDURLSessionProvider request-isolation issue is not claimed fixed by these changes.


OskarEichler:
Superseded by focused PRs so each independently reviewable root cause has its own patch:

- #1138 — stale custom-header reuse
- #1139 — concurrent interactive request-state mixing
- #1140 — blocking prefetch coordination

The combined branch is preserved, but no longer needs review.

PR #1143, opened 2026-09-04
[LOW] security(ci): pin actions to immutable revisions

OskarEichler:
## Security impact

The pull-request and package-release workflows execute actions/checkout@v4, actions/setup-node@v4, and changesets/action@v1 through mutable major-version references. If one of those references is moved or its publishing account is compromised, unreviewed code can execute in CI. The release job has repository write access and npm trusted-publishing access, so immutable action identity matters at that boundary.

## Fix

Pin every external action invocation to the exact commit currently selected by its existing major reference, retaining a version comment for update tooling and reviewers:

- actions/checkout v4.4.0
- actions/setup-node v4.4.0
- changesets/action v1.9.0

This does not upgrade action majors or change workflow behavior. Token permissions and the floating npm install are separate focused changes.

## Verification

- Both workflow files parse successfully as YAML
- Every uses reference is an immutable 40-character commit SHA
- Each SHA was cross-checked against the named upstream release/tag
- git diff --check passes

No runtime package changes or changeset are needed for CI-only hardening.


.github/workflows/pull-request.yml
```diff
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
```
ilteoood, line 12: At this point it would be worth to upgrade them to their latest version

PR #1145, opened 2026-09-04
[LOW] security(ci): pin the release npm version

OskarEichler:
## Security impact

The privileged release job installs npm@^11.5.1 immediately before publishing. That range resolves again on every release, so a newly published npm 11.x version can execute with repository write and npm OIDC privileges before this repository has reviewed it.

## Fix

Pin the install to npm 11.19.1, the current npm 11 release. npm 12 is not selected because this workflow runs Node 20 and npm 12 requires a newer Node line; npm 11.19.1 supports Node ^20.17.0 or >=22.9.0 and retains the required trusted-publishing support.

This PR changes only the floating release-tool version. Action revision pins and workflow permissions are separate focused changes.

## Verification

- npm 11.19.1 resolves and reports the expected version
- Package engine compatibility was checked against the workflow's Node 20 runtime
- release.yml parses successfully as YAML
- git diff --check passes

No runtime package changes or changeset are needed for CI-only hardening.


.github/workflows/release.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           cache: 'yarn'
 
       - name: Upgrade npm for OIDC
-        run: npm i -g npm@^11.5.1
+        run: npm i -g npm@11.19.1
```
ilteoood, line 30: At this point, installing the latest 11 version would be more than enough

```suggestion
        run: npm i -g npm@11
```

PR #1146, opened 2026-09-23
feat: allow to resume the request if androdi OS kills the process

Francesco-Voto:
# Description

On Android, the OS can kill the app process while the user is away in the Custom Tab completing
login (e.g. under memory pressure). When the app restarts, React Native drops the resulting
activity result because it arrives before the JS context is ready
(facebook/react-native#30277), so the in-flight `authorize()` promise is lost and the user's
login silently fails to complete even though the browser already returned a valid response.

This adds a `resumePendingAuthorize()` API to recover from that case:

- `RNAppAuthModule` now stashes the raw activity result in a static field
  (`stashAuthorizationResult`) when it's received with no in-flight `authorize()` promise on the
  module instance, instead of discarding it.
- The new `resumePendingAuthorize` native method and JS/TS export claim that stashed result,
  re-derive the token request from the `AuthorizationResponse` (which carries the PKCE code
  verifier), and complete the token exchange.
- Consumers must forward `MainActivity.onActivityResult` to
  `RNAppAuthModule.stashAuthorizationResult` and call `resumePendingAuthorize()` on startup;
  it resolves `null` (safe to call unconditionally) when there's nothing to resume, and always
  resolves `null` on iOS.
- Also moves `AuthorizationService` construction and `createCustomTabsIntentBuilder()` off the
  main thread in `authorize()`, since both are `@WorkerThread` calls that could block the UI for
  up to a second.
- Documented the new API and the required `MainActivity` wiring in
  `docs/docs/usage/authorization.md`.

## Steps to verify

1. Wire up `MainActivity.onActivityResult` to call `RNAppAuthModule.stashAuthorizationResult`
   as shown in the updated docs, and call `resumePendingAuthorize()` on app startup.
2. Run `authorize()` against a test IdP, and while the Custom Tab is open, kill the app process
   from Android Studio's Logcat ("Terminate Application") or via
   `adb shell am kill <package>`.
3. Complete the login in the browser; the app should cold-start, and `resumePendingAuthorize()`
   should resolve with a valid token response instead of the flow silently failing.
4. Confirm a normal (non-killed) `authorize()` flow still resolves as before, and that
   `resumePendingAuthorize()` resolves `null` when called with nothing pending.
5. Confirm `resumePendingAuthorize()` resolves `null` immediately on iOS.


packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -65,11 +65,20 @@
...
 
 public class RNAppAuthModule extends ReactContextBaseJavaModule implements ActivityEventListener {
 
     public static final String CUSTOM_TAB_PACKAGE_NAME = "com.android.chrome";
 
+    public static final int AUTHORIZATION_REQUEST_CODE = 52;
+
+    private static final AtomicReference<Intent> sStashedAuthorizationResult = new AtomicReference<>();
+
+    public static void stashAuthorizationResult(Intent data) {
```
ilteoood:
```suggestion
    public static void stashAuthorizationResult(final Intent data) {
```

packages/react-native-app-auth/android/src/main/java/com/rnappauth/RNAppAuthModule.java
```diff
@@ -484,13 +493,73 @@ public void onFetchConfigurationCompleted(
...
+            final AuthorizationService authService = new AuthorizationService(this.reactContext, configuration);
+
+            final Map<String, String> additionalParametersMap = MapUtil.readableMapToHashMap(additionalParameters);
+            final TokenRequest tokenRequest = additionalParametersMap.isEmpty()
+                    ? response.createTokenExchangeRequest()
+                    : response.createTokenExchangeRequest(additionalParametersMap);
+
+            authService.performTokenRequest(tokenRequest, new AuthorizationService.TokenResponseCallback() {
+                @Override
+                public void onTokenRequestCompleted(TokenResponse resp, AuthorizationException ex) {
```
ilteoood:
```suggestion
                public void onTokenRequestCompleted(final TokenResponse resp, final AuthorizationException ex) {
```

packages/react-native-app-auth/index.js
```diff
@@ -274,6 +274,29 @@ export const authorize = ({
...
+    return Promise.resolve(null);
+  }
+
+  validateHeaders(customHeaders);
+  validateConnectionTimeoutSeconds(connectionTimeoutSeconds);
+
+  return wrapNativeAuthPromise(
+    RNAppAuth.resumePendingAuthorize(
+      additionalParameters,
+      convertTimeoutForPlatform(Platform.OS, connectionTimeoutSeconds),
```
ilteoood:
```suggestion
} = {}) => {
  const platform = Platform.OS
  if (platform !== 'android') {
    return Promise.resolve(null);
  }

  validateHeaders(customHeaders);
  validateConnectionTimeoutSeconds(connectionTimeoutSeconds);

  return wrapNativeAuthPromise(
    RNAppAuth.resumePendingAuthorize(
      additionalParameters,
      convertTimeoutForPlatform(platform, connectionTimeoutSeconds),
```

